PR #5, opened 2024-06-27
Add initial POC code for javascript instrumentation

kshitiz-okahu:
The code contains instrumentation POC code for LangChain and LlamaIndex frameworks.
The code uses [commonJS modules](https://nodejs.org/api/modules.html).
The POC for [Modules: ECMAScript](https://nodejs.org/api/esm.html#modules-ecmascript-modules) will be added later


oi-raanne:
My recommendation would be to maintain separate repository for each programming language, as it would ease the management of continuous integration with code and security checks in Github, and release process for committers  would be without any ambiguity.

prasad-okahu:
> My recommendation would be to maintain separate repository for each programming language, as it would ease the management of continuous integration with code and security checks in Github, and release process for committers would be without any ambiguity.

@oi-raanne , agree with the proposition. Let's keep the current repo for python and create a new one for typescript. In future, we can follow this practice for any new language that we add support for.
I'll go ahead create the new repo. Thanks

Issue #25, opened 2024-08-08
Look into the naming of workflow_context_properties 


kshitiz-okahu:
The proposal is to decide on naming of attribute for propagation of workflow context across GenAI traces.

kshitiz-okahu:
The consensus is that the naming should not include any genAI references.

I am in favour of naming these fields as follows

context.session_id

oi-raanne:
'invocation' would be better than 'context', as 'context' is first class object in open telemetry

PR #34, opened 2024-08-29
Update key for session context field in attributes

kshitiz-okahu:
## Proposed changes

The key for persisting sessions in attributes has been simplified as discussed in https://github.com/monocle2ai/monocle/issues/25

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/tests/langchain_chat_sample.py
+++ b/tests/langchain_chat_sample.py
@@ -14,7 +14,7 @@ from langchain_openai import ChatOpenAI, OpenAIEmbeddings
 from langchain_text_splitters import RecursiveCharacterTextSplitter
 from monocle_apptrace.instrumentor import set_context_properties, setup_monocle_telemetry
 from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
-from temp_langchain import create_history_aware_retriever
+from langhchain_patch import create_history_aware_retriever
 
 setup_monocle_telemetry(
             workflow_name="langchain_app_1",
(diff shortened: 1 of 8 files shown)


tests/langchain_chat_sample.py
```diff
@@ -154,7 +154,7 @@ def format_docs(docs):
 #         "status_code": "UNSET"
 #     },
 #     "attributes": {
-#         "workflow_context_properties.session_id": "0x4fa6d91d1f2a4bdbb7a1287d90ec4a16",
+#         "context.session_id": "0x4fa6d91d1f2a4bdbb7a1287d90ec4a16",
 #         "workflow_context_input": "What is Task Decomposition?"
```
oi-raanne, line 158: input is no longer captured in attributes

tests/langchain_chat_sample.py
```diff
@@ -154,7 +154,7 @@ def format_docs(docs):
 #         "status_code": "UNSET"
 #     },
 #     "attributes": {
-#         "workflow_context_properties.session_id": "0x4fa6d91d1f2a4bdbb7a1287d90ec4a16",
+#         "context.session_id": "0x4fa6d91d1f2a4bdbb7a1287d90ec4a16",
```
oi-raanne, line 157: as context is a first class object in telemetry, it can confuse the users as json path to fetch would be context.<property name>.  My suggestion would be to use either 'invocation' or not to add any prefix

oi-raanne (changes requested):

tests/langchain_chat_sample.py
akshay-okahu: - from perspective of capturing and querying these attributes in prompt API, should be fine either way
  - if the span of these attributes is fixed, like root span for eg, it would be easier for prompt API implementation
  - since these attributes are inserted in all spans, including root span, so prompt API implementation should be fine
- some prefix might be helpful to determine if the attribute came from framework vs custom attribute inserted by user, though not sure if this is needed
  -  some possible prefix alternatives could be: state, session, trace

kshitiz-okahu: changed to session.{property_name}

oi-raanne (approved):

akshay-okahu (approved):

PR #311, opened 2025-10-17
Support skiping entity execution and mock response.

prasad-okahu:
## Proposed changes

- Add ability in Monocle apptrace to skip underlying instrumented API and return a preset response
  - Support placeholders in the preset response (eg a response from tool call can be formatted using tool input args)
- Allow skipping the API for a specific entity name and type eg skip hotel_tool only and not flight and weather tool
- Support setting the preset from the test framework
- The span formatting is now split in pre-execution and post execution so that we can make the skip execution decision based on the entities and input.

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments


apptrace/src/monocle_apptrace/instrumentation/common/span_handler.py
```diff
@@ -306,6 +320,32 @@ def workflow_type(to_wrap=None, span:Span=None):
...
+    def get_iput_entity_type(span: Span) -> str:
+        for event in span.events:
+            if event.name == "data.input":
+                return event.attributes.get("entity.type", "")
+
+    @staticmethod
+    def skip_execution(span:Span) -> tuple[bool, Union[dict, list, str, None]]:
+        skip_execs = get_value(MONOCLE_SKIP_EXECUTIONS)
+        if skip_execs is not None:
+            skip_exec_entity = skip_execs.get(span.attributes.get("entity.1.name", ""),{})
```
oi-raanne, line 333: why we explicitly checking for entity.1.name or entity.1.type ?

apptrace/src/monocle_apptrace/instrumentation/common/span_handler.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import os
```
oi-raanne, line 2: can't there be a test_span_handler.py and use the same for testing purpose rather than changing the existing one ?

apptrace/src/monocle_apptrace/instrumentation/common/span_handler.py
prasad-okahu: The objective is to skip a specific entity ie skip execution of an agent 'travel-agent' and not all agent executions. Hence the name and type check.

apptrace/src/monocle_apptrace/instrumentation/common/span_handler.py
prasad-okahu: This is not just a test feature. The overall idea is to inject pre and post execution. Mocking for test is first use case here. Beyond that it can support things like guardrails, memory injection etc. Hence it's part of Monocle core. Also, we override the span handler in many meta models for various features. Using a separate test handler will be problematic.

oi-raanne: This looks way complex and may be designed better. Also, somewhere we are progressing with test tool related changes inside monocle src without design discussion.

oi-raanne (changes requested):
mock response generation should not be part of monocle src code

oi-raanne (approved):

PR #332, opened 2025-11-14
Added the scope session_id for adk

sachintendulkar576123:
## Proposed changes

This PR:

- Added the scope session_id for adk

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/metamodel/anthropic/_helper.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/metamodel/anthropic/_helper.py
@@ -17,6 +17,7 @@ from monocle_apptrace.instrumentation.common.utils import (
 )
 from monocle_apptrace.instrumentation.metamodel.finish_types import map_anthropic_finish_reason_to_finish_type
 from monocle_apptrace.instrumentation.common.constants import AGENT_PREFIX_KEY, INFERENCE_AGENT_DELEGATION, INFERENCE_TURN_END, INFERENCE_TOOL_CALL
+from contextlib import suppress
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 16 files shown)


apptrace/src/monocle_apptrace/instrumentation/metamodel/adk/_helper.py
```diff
@@ -207,3 +209,15 @@ def get_target_agent(instance: Any) -> str:
         str: The name of the target agent
     """
     return getattr(instance, 'name', getattr(instance, '__name__', 'unknown_target_agent'))
+
+
+class AdkSpanHandler(SpanHandler):
```
prasad-okahu, line 215: @sachintendulkar576123 , can we generalize this? Basically we'll need this logic in every agentic handler. Perhaps, we should have a common handler base class that sets this scope in pre_tracing and ends it in post_tracing. The subclass can just set the session id class member and call the super()'s pre_tracing.
Also, the scope name should be consistent across all agent instrumentation, let's use "agentic.session_id"

apptrace/src/monocle_apptrace/instrumentation/metamodel/adk/_helper.py
```diff
@@ -207,3 +209,15 @@ def get_target_agent(instance: Any) -> str:
...
+    """Custom span handler for ADK instrumentation that adds session_id scope."""
+
+    def pre_tracing(self, to_wrap, wrapped, instance, args, kwargs):
+        """Set session_id scope before tracing begins."""
+        session_id = kwargs.get('session_id')
+        if session_id:
+            set_scope('adk.session_id', session_id)
+
+        return super().pre_tracing(to_wrap, wrapped, instance, args, kwargs)
```
prasad-okahu: you should return the token created by scope instead of returning the super().pre_tracing(). Also, pre_tracing returns a tuple where the second value is an alternate wrapper. Since that not applicable here, it should be 
`return set_scope(AGENT_SESSION, session_id), None`

apptrace/src/monocle_apptrace/instrumentation/metamodel/adk/_helper.py
prasad-okahu: I guess on second thought, it'll be a bit messy to create that base class. I am fine with the current approach. Just set a   `AGENT_SESSION = "agentic.session"` in constants.py so it can used by all handlers.

apptrace/src/monocle_apptrace/instrumentation/metamodel/adk/_helper.py
```diff
@@ -207,3 +210,32 @@ def get_target_agent(instance: Any) -> str:
         str: The name of the target agent
     """
     return getattr(instance, 'name', getattr(instance, '__name__', 'unknown_target_agent'))
+
+
+class AdkSpanHandler(SpanHandler):
+    """Custom span handler for ADK instrumentation that adds session_id scope."""
+
+    def pre_tracing(self, to_wrap, wrapped, instance, args, kwargs):
```
prasad-okahu, line 218: 1. I suggest you skip the super().pre_trace(). Just start the scope and return the token, null. The base class pre_tracing() is no-op.
2. The default post_tracing already handles detaching the token which takes care of removing the scope. You don't need the post_tracing() code here.

prasad-okahu (approved):

PR #529, opened 2026-04-27
Claude code instrumentation

imohammedansari:
## Proposed changes

Adds Monocle instrumentation support for Claude Code CLI. Uses Claude Code's built-in hook system to automatically trace sessions — tool calls, subagent invocations, and emit them as OpenTelemetry spans.                                                                                                                  
   
## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments


apptrace/src/monocle_apptrace/__main__.py
```diff
@@ -2,10 +2,28 @@
...
 def main():
+    if len(sys.argv) >= 2:
+        cmd = sys.argv[1]
+
+        if cmd in ("claude-hook", "claude_hook", "--claude-hook"):
+            from monocle_apptrace.instrumentation.metamodel.claude_cli.event_handler import main as hook_main
+            hook_main()
+            sys.exit(0)
+
+        if cmd in ("claude-install", "claude_install"):
```
prasad-okahu: I think we should call it 'claude-setup'

apptrace/src/monocle_apptrace/instrumentation/metamodel/claude_cli/hooks.json
```diff
@@ -0,0 +1,125 @@
+{
```
prasad-okahu, line 1: Every ' python -m monocle_apptrace' command will require the python virtual env sourced with monocle_apptrace installed. How do we ensure that? Do you have to start Claude with the env sourced? Should we source the vent from the hook command instead?

prasad-okahu (commented):
Overall look fine to me. A couple of suggestions below. Also, I think we'll need to handle the tool error using the post tool failure hook. Please check that. Feel free to include that change here or a follow up patch.

apptrace/src/monocle_apptrace/__main__.py
imohammedansari: Changed the name

apptrace/src/monocle_apptrace/instrumentation/metamodel/claude_cli/hooks.json
imohammedansari: I made it so we don’t depend on the user activating the virtual environment when Claude runs.
During claude-setup, we replace python with the full path to the Python used for setup.

So instead of:
python -m monocle_apptrace...
 #It saves:
/path/to/venv/bin/python3 -m monocle_apptrace...

When Claude runs the hook, it uses that exact Python directly, so no activation is needed.
The only requirement is that setup is run from the correct environment.

prasad-okahu (approved):

PR #538, opened 2026-04-30
Codex instrumentation and UV packaging 

imohammedansari:
## Proposed changes

Adds Monocle instrumentation support for Codex CLI. Uses Codex Code's built-in hook system to automatically trace sessions — tool calls, subagent invocations, and emit them as OpenTelemetry spans.

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments

Next steps: Better packaging for both Codex and Claude CLI instrumentation


apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
```diff
@@ -0,0 +1,175 @@
...
+REQUEST = {
+    "type": SPAN_TYPES.AGENTIC_REQUEST,
+    "subtype": SPAN_SUBTYPES.TURN,
+    "attributes": [
+        [
+            {"attribute": "type", "accessor": lambda arguments: "agent.codex_cli"},
+            {"attribute": "name", "accessor": lambda arguments: "Codex CLI"},
+            {"attribute": "turn_id", "accessor": lambda arguments: arguments["kwargs"].get("turn_id", "")},
+            {"attribute": "time_to_first_token_ms", "accessor": lambda arguments: arguments["kwargs"].get("time_to_first_token_ms")},
+            {"attribute": "duration_ms", "accessor": lambda arguments: arguments["kwargs"].get("duration_ms")},
```
prasad-okahu: Is the duration getting captured via the start time and stop time on the span itself? If that's the case, we probably don't need the duration here.

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
```diff
@@ -0,0 +1,175 @@
...
+}
+
+REQUEST = {
+    "type": SPAN_TYPES.AGENTIC_REQUEST,
+    "subtype": SPAN_SUBTYPES.TURN,
+    "attributes": [
+        [
+            {"attribute": "type", "accessor": lambda arguments: "agent.codex_cli"},
+            {"attribute": "name", "accessor": lambda arguments: "Codex CLI"},
+            {"attribute": "turn_id", "accessor": lambda arguments: arguments["kwargs"].get("turn_id", "")},
```
prasad-okahu: Is this a unique turn ID generated by codex? If it is, can set this as turn ID scope instead of generating one by normal monocle agentic span handling?

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    "attribute": "response",
+                    "accessor": lambda arguments: arguments["kwargs"].get("output_text", ""),
+                }
+            ],
+        },
+        {
+            "name": "metadata",
+            "attributes": [
+                {"accessor": lambda arguments: arguments["kwargs"].get("tokens", {})},
```
prasad-okahu, line 53: Does this cover input, output and total token?

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
```diff
@@ -0,0 +1,175 @@
...
+            {
+                "attribute": "name",
+                "accessor": lambda arguments: arguments["kwargs"].get("agent_nickname") or arguments["kwargs"].get("agent_role", "agent"),
+            },
+            {
+                "attribute": "role",
+                "accessor": lambda arguments: arguments["kwargs"].get("agent_role", ""),
+            },
+            {
+                "attribute": "thread_id",
```
prasad-okahu: Is the thread ID same as agent invocation ID ie unique for each execution of this agent? If it is, can we use this as Monocle agent invocation ID scope?

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/tool.py
```diff
@@ -0,0 +1,75 @@
...
+
+TOOL = {
+    "type": SPAN_TYPES.AGENTIC_TOOL_INVOCATION,
+    "subtype": SPAN_SUBTYPES.CONTENT_GENERATION,
+    "attributes": [
+        [
+            {"attribute": "type", "accessor": lambda arguments: _helper.get_tool_type(arguments)},
+            {"attribute": "name", "accessor": lambda arguments: _helper.get_tool_name(arguments)},
+            {"attribute": "description", "accessor": lambda arguments: _helper.get_tool_description(arguments)},
+            {"attribute": "call_id", "accessor": lambda arguments: arguments["kwargs"].get("call_id", "")},
```
prasad-okahu: Is this a unique tool invocation ID generated by codex?

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/tool.py
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    "attribute": "response",
+                    "accessor": lambda arguments: _helper.extract_tool_response(arguments["result"]),
+                }
+            ],
+        },
+    ],
+}
+
+MCP_TOOL = {
```
prasad-okahu, line 40: Do we capture MCP server information? I believe there's separate span that captures the MCP server metadata (name, endpoint etc).

prasad-okahu (commented):
Looks fine. A few comments/question below.

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
imohammedansari: Yeah, it’s already captured in the timestamps, so this is redundant.

Also, thinking about it, time_to_first_token_ms is redundant too. The idea was to measure latency from when the first answer starts to when the first token is generated, but it’s not really that useful.

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
imohammedansari: Yes, it is. I updated it to scope the turn ID the same way we do for the session ID.

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
imohammedansari: Yes it does, it includes prompt_tokens, completion_tokens, total_tokens, cache_read_tokens

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/agent.py
imohammedansari: Yes it is, I updated it to be used as the agent invocation ID scope

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/tool.py
imohammedansari: Yes, it is. Should I add an attribute like scope.agentic.tool.invocation and set that to this?

apptrace/src/monocle_apptrace/instrumentation/metamodel/codex_cli/entities/tool.py
imohammedansari: The transcripts don’t include the server details for mcp tool calls. They only include the tool name. We already added the mcp name, but the transcript does not provide the server URL.

apptrace/src/monocle_apptrace/instrumentation/common/utils.py
imohammedansari, line 1: @prasad-okahu 
To get the turn ID and invocation ID to match the ones generated by Codex, I had to make a small change.

What was happening is that when we set these values, they were getting overwritten by a new random value during span setup. To fix this, I added a simple check before generating a new value, it checks if something already exists for that scope. If it does, it keeps that instead of overwriting it.

What do you think? Could this break anything in other frameworks where we may actually want to override the existing ID?

prasad-okahu: Yes, I think this would be a problem for normal tracing. For example, a root agent invocation starts and gets it own invocation ID. Then it invokes a child agent which should get its own invocation ID. However with this change, the parent's invocation ID will continue to be used for child.
I guess it's okay to not reuse codex generated IDs in place of Monocle generated ones. Perhaps we capture them with a different scope name just to track.

prasad-okahu (commented):
Looks fine, just one feedback on replacing Monocle IDs with Codex

apptrace/src/monocle_apptrace/exporters/okahu/okahu_exporter.py
```diff
@@ -45,11 +45,13 @@ def _get_monocle_env_value(key: str) -> Optional[str]:
     if env_value:
         return env_value
 
-    # Support both filenames for compatibility.
-    env_file_names = (".env.monocle",)
+    # Lookup order: ./env.monocle (per-project override) → ~/.monocle/.env (global).
```
prasad-okahu: We might have to move .env.monocle file under .monocle folder in next version. The .monocle will have env, traces and config folder to store different items. 
Can you move the file handling logic under a common method in util?

apptrace/src/monocle_apptrace/instrumentation/metamodel/claude_cli/hooks.json
prasad-okahu, line 1: I guess the python env/path will be handled separately if needed.

prasad-okahu (commented):
Looks fine, just one change requested for the file path

apptrace/src/monocle_apptrace/instrumentation/metamodel/claude_cli/hooks.json
imohammedansari: When we run uv tool install monocle-apptrace, it does creates its own isolated environment and installs monocle-apptrace

After that, whenever we run monocle-apptrace codex-hook or monocle-apptrace claude-hook, it always uses the Python and dependencies from that uv-managed environment.

Even if we activate a project venv and then run monocle-apptrace, it does not use the .venv Python; it still uses the environment created by the uv.

So no changes are required. I tested both Codex and Claude in an isolated VM, and the installation works properly and all traces are being sent to Okahu correctly.

prasad-okahu (approved):

PR #586, opened 2026-05-25
Copilot CLI and VS Code Copilot Instrumentation

imohammedansari:
## Proposed changes
Adds Monocle instrumentation support for GitHub Copilot, including Copilot CLI and VS Code Copilot Chat agent mode. Uses Copilot's built-in hook system to automatically trace sessions, tool calls, subagent invocations, model used, and per-session token consumption for Copilot CLI, then emits them as monocle telemetry.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/monocle2ai/monocle/blob/main/CONTRIBUTING.md) doc
- [ ] I have signed the CLA
- [X] I have added the necessary documentation (if appropriate)

## Further comments
Token consumption is captured only for Copilot CLI, not VS Code Copilot Chat (VS Code doesn't expose tokens). Tokens are reported per session and only emitted when the session ends.

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/metamodel/github_copilot/replay.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/metamodel/github_copilot/replay.py
@@ -395,46 +395,6 @@ def _build_subagent_records(interaction: dict, subagent_outputs_by_id: dict) -> 
     return records
 
 
-def _normalize_tokens(usage: dict) -> dict:
-    """Canonicalize Copilot CLI's modelMetrics.usage. prompt_tokens sums
-    input+cacheRead+cacheWrite — raw inputTokens alone undercounts cached
-    prompts on warm sessions."""
-    input_t = usage.get("inputTokens", 0) or 0
-    output_t = usage.get("outputTokens", 0) or 0
-    cache_read = usage.get("cacheReadTokens", 0) or 0
-    cache_write = usage.get("cacheWriteTokens", 0) or 0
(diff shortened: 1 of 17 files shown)


apptrace/src/monocle_apptrace/instrumentation/metamodel/github_copilot/replay.py
```diff
@@ -0,0 +1,564 @@
...
+                "agent_type": itv["agent_type"],
+                "prompt": "",
+                "assistant_messages": [],
+                "tool_calls": [],
+                "model": current_model,
+                "start_ts": itv["start_ts"],
+                "end_ts": itv["end_ts"],
+            }
+            current.setdefault("subagents", []).append(current_subagent)
+        return current_subagent, True
```
prasad-okahu: Nit: the second return value doesn't seem to be used anywhere

prasad-okahu (commented):
@imohammedansari , can we add unit tests that generate span using the captured raw hook output?

apptrace/src/monocle_apptrace/instrumentation/metamodel/github_copilot/replay.py
imohammedansari: Removed the second boolean value

imohammedansari:
> can we add unit tests that generate span using the captured raw hook output?

Yes, we should be able to. I can create unit tests for the instrumentations in a follow-up PR.

prasad-okahu (approved):

PR #666, opened 2026-06-26
strands: anchor agentic.session from Agent.trace_attributes

imohammedansari:
## Problem

Strands `Agent`s commonly carry their session id in `Agent.trace_attributes` (the standard Strands mechanism, e.g. `Agent(trace_attributes={"session.id": ...})`) rather than via a `SessionManager`. Monocle's `extract_session_id` only checked `_session_manager`, so any Strands agent without a session manager emitted **no `scope.agentic.session`** — leaving its traces unanchored.

## Reproduction

Instrumented a real Strands app (DevDuck) running on OpenAI. The trace was structurally correct (`workflow → agentic.turn → agentic.invocation → {inference, agentic.tool.invocation}`, all I/O populated, tokens + model present), but every span had an empty `scope.agentic.session` because DevDuck attaches its session via `trace_attributes={"session.id": ...}` and uses no `SessionManager`.

## Fix

`extract_session_id` falls back to `instance.trace_attributes` (`session.id`, then `session_id`) when no `session_manager` is present. After the fix, all spans in the trace share the same `scope.agentic.session`.

## Test

Adds `test_strands_handler_uses_trace_attributes_session` covering the fallback. Existing session tests unchanged and still pass (the no-session case still yields no scope).

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/metamodel/strands/_helper.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/metamodel/strands/_helper.py
@@ -27,6 +27,13 @@ def extract_session_id(instance,kwargs):
     if "session_manager" in kwargs and hasattr(kwargs["session_manager"],'session_id'):
         return kwargs["session_manager"].session_id
 
+    # Agents without a session_manager commonly carry a session id in trace_attributes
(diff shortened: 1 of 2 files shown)


imohammedansari:
Reopening from correctly-named branch mohammed/strands-trace-attributes-session.

PR #682, opened 2026-07-05
Samiulmushfik/postgres exporter

### LINKED ISSUE #604: Support Postgres exporter
Currently Monocle supports exporting to file, cloud store (S3, blob, GFS etc). We want to add an ability to store the traces into a Postgres database.
This ticket is tracking the ask to create a Postgres exporter that write the traces to a Postgres database.

## Configuration
The exporter should accept the database specific configuration from following env variables
- MONOCLE_POSTGRES_CONNECTION_URL : Postgres connection URL

##Database schema
 The spans from traces should be stored in a table called 'traces' with following schema
Column | PostgreSQL Type | Span field| Details
-- | -- | -- | --
name | TEXT | name | Span name
start_time | TIMESTAMP WITH TIME ZONE| start_time | Span start time
end_time | TIMESTAMP WITH TIME ZONE| end_time | Span end time
status | JSONB | status | Span status that includes Status.State and State.message
span_id | VARCHAR(64)  | context.span_id | Span ID in HEX format with leading 0x
trace_id | VARCHAR(64)  |  context.trace_id |Trace ID in HEX format with leading 0x
parent_id | VARCHAR(64)  | parent_id| Span ID of parent span in HEX format with leading 0x, NULL if there's no parent span
attributes | JSONB | attributes | attributes json
events | JSONB | events | event json
metadata | JSONB | - | optional metadata for the exporter or the collector to add
insertion_timestamp | TIMESTAMP WITH TIME ZONE | - | timestamp - when the record was added to table (default CURRENT_TIMESTAMP)

## Pre-requisite 
- The traces table should be created in the database/schema where the connection URL is pointing.
- The user specified in the connection URL should have INSERT permission for the traces table

## How to export
The exporter should extract the information from each span and insert the record into the table.

samiul123:
## Proposed changes

This PR adds a PostgreSQL span exporter (PostgresSpanExporter) that writes OpenTelemetry traces produced by Monocle to `traces` table in a PostgreSQL database. Users only need to set `MONOCLE_EXPORTER`=postgres and `MONOCLE_POSTGRES_CONNECTION_URL`.

Key behaviors:

- Spans are inserted as a single batch directly to the DB
- The traces table is auto-created on startup via CREATE TABLE IF NOT EXISTS; raises PermissionError with a clear message if the DB user lacks CREATE TABLE privilege
- Connection failures trigger a single reconnect-and-retry before failing

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments

The exporter is registered in monocle_exporters under the key `postgres` so it integrates with the existing exporter selection mechanism. The integration test requires a running Postgres instance and is automatically skipped when `MONOCLE_POSTGRES_CONNECTION_URL` is not set.


samiul123:
Hi @prasad-okahu, Let me know if I need to change anything.

prasad-okahu:
@samiul123 Thanks for your contribution!
@shambhavigoel-13 could you also please take a look and provide feedback.

apptrace/src/monocle_apptrace/exporters/postgres/postgres_exporter.py
```diff
@@ -0,0 +1,194 @@
...
+                self._do_insert(rows)
+            except psycopg2.OperationalError as e:
+                logger.warning("DB connection error, attempting reconnect: %s", e)
+                self._reconnect()
+                self._do_insert(rows)
+        del self.trace_spans[trace_id]
+
+    def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
+        try:
+            self._cleanup_expired_traces()
```
prasad-okahu: I don't think we need to handle the complexity of combining multiple spans of a single trace in the exporter. The file exporter does that so it can keep all the spans belonging to a trace together in a single file. This way your can retrieve the full trace as one file. For a database store, you can simply query by the traceID for that.
I would suggest making the trace export a simple batch insert into the trace table.

prasad-okahu (commented):
@samiul123 Again, thanks for the contribution. The changes overall look fine. My main feedback is about sorting spans for a single trace in the exporter. Please see the details in the review.

apptrace/src/monocle_apptrace/exporters/postgres/postgres_exporter.py
samiul123: Thanks for the feedback @prasad-okahu . Will change accordingly.

samiul123: I updated the logic @prasad-okahu . Please have a look. Thanks.

prasad-okahu (approved):

prasad-okahu:
@samiul123 Merged the patch. 
Thanks for your contribution! Glad to have you as part of Monocle community!

samiul123:
Thank you, @prasad-okahu. I really enjoyed it. I want to work more deeply on this and keep on contributing.

PR #708, opened 2026-07-11
Add OpenHands agent SDK instrumentation

imohammedansari:
## Proposed changes

Adds instrumentation support for **OpenHands** (`openhands-sdk` — the [software-agent-sdk](https://github.com/OpenHands/software-agent-sdk) that powers OpenHands v1, its CLI, and OpenHands Cloud). OpenHands is a custom agent loop (no LangGraph/CrewAI underneath), so today a single `setup_monocle_telemetry()` call produces only disconnected per-LLM-call traces. This PR adds the framework metamodel so the same one call produces the full agentic span tree:

```
workflow
└── agentic.turn            LocalConversation.run / arun
    └── agentic.invocation  Agent.step / astep        (1..N)
        ├── inference       (existing litellm metamodel — no litellm changes in this PR)
        └── agentic.tool.invocation  ParallelToolExecutor._run_safe / _arun_safe
            ├── agentic.mcp.invocation      (existing mcp metamodel, now correctly parented)
            └── agentic.turn → …            (delegated sub-conversations, nested)
```

The change is purely additive: a new `metamodel/openhands/` package, its registry entries, and a debug-level version-compat guard. No shared behavior is modified; apps without OpenHands installed are unaffected (import-guarded like every framework).

## Changes

1. **New `metamodel/openhands/` package**
   - `methods.py`: `LocalConversation.run/arun` → `agentic.turn`; `Agent.step/astep` → `agentic.invocation`; `ParallelToolExecutor._run_safe/_arun_safe` → `agentic.tool.invocation`; `execute_batch` as a stash-only (never a span) boundary.
   - `entities/agent.py`, `entities/tool.py`: attribute/event accessors over OpenHands' typed Pydantic events (`ActionEvent` in, `ObservationEvent`/`AgentErrorEvent` out); turn/step I/O read from the event-sourced conversation state.
   - `_helper.py`: pure duck-typed accessors — never imports `openhands` (the registry imports every metamodel at setup time for every user).
   - `openhands_processor.py`:
     - `OpenHandsSpanHandler` anchors `scope.agentic.session` to the stable conversation id; delegated sub-conversations inherit the parent session instead of starting their own.
     - `OpenHandsToolHandler` does per-action **context stitching**: OpenHands executes tools on worker threads (`run_in_executor` async, `ThreadPoolExecutor` when `tool_concurrency_limit > 1`), which drop contextvars — naive wrapping produced orphan single-span traces. The handler stashes the OTEL context per action event at the caller-side boundary and restores it thread-side, so tool spans, MCP invocations, and delegated sub-conversations parent correctly in every execution mode (sync/async, sequential/parallel). A context-guarded `skip_span` drops the duplicate thread-side inner call.

2. **Registry** (`common/wrapper_method.py`): `OPENHANDS_METHODS` added to `DEFAULT_METHODS_LIST`; `openhands_handler` / `openhands_tool_handler` added to `MONOCLE_SPAN_HANDLERS`.

3. **Version compatibility** (`common/instrumentor.py`): `openhands-sdk` 1.21 (pinned by the OpenHands CLI) predates the async API (`arun`/`astep`/`_arun_safe`); those methods get the same debug-level version-compat handling that `agents.run` already uses. Sync path is fully covered on 1.21.

4. **Tests**: `tests/unit/test_openhands_metamodel.py` — 12 tests with stand-in classes (no OpenHands install needed): session anchoring + inheritance, context stash/restore across the thread hop, duplicate/orphan skip rules, all I/O accessors, nested-turn skip.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

**Validated with real runs** (`gpt-4o-mini`, openhands-sdk 1.35 and the 1.21 CLI): sync single-turn, async multi-turn (shared session), MCP tools sync + async (previously orphaned in async), agent delegation with parallel task tools (previously three disconnected traces), and the headless OpenHands CLI — every trace passes the full recipe with non-empty `data.input`/`data.output` on all agentic spans. `tests/unit`: all failures present are byte-identical on clean `main` (pre-existing environment issues, verified by stash/re-run) — zero regression.

**Why the tool boundary is `ParallelToolExecutor` and not `Agent._execute_action_event`:** the innermost per-tool method runs on a worker thread with no OTEL context; `_arun_safe` is a coroutine on the event-loop thread (context intact even under `asyncio.gather`), and `_run_safe` runs on the calling thread in the default sequential config. Context stitching covers the remaining thread-hop cases rather than dropping spans.

**Scope note:** LLM inference spans come from the existing litellm metamodel unchanged. litellm-level streaming/Responses-API coverage (which OpenHands also exercises) is deliberately split into a separate PR so this one stays isolated to OpenHands. Remote/sandboxed agent-server mode is cross-process and planned as an upstream OpenHands integration.

**Review round (adversarial review applied):** removed dead `should_skip` wiring on the turn entity (single-`output_processor` configs never consult it — nested delegated turns are intentional, validated trace shape); context stash now keyed by the SDK's stable `ActionEvent.id` (object-identity keys could theoretically re-parent after GC reuse); duplicate-skip now matches the exact stashed action instead of any tool-span parent (protects tool-in-tool); `_get_events` is a lazy newest-first generator (EventLog indexing can hit disk per event on persisted conversations); turn output prefers a definitive outcome (agent message / finish message / error) over a trailing action summary; the instrumentor version-compat guard is narrowed to `AttributeError`; stash lifecycle and duplicate-skip are unit-tested with an isolating fixture. Re-validated: 5 live scenarios PASS with zero I/O misses.


apptrace/src/monocle_apptrace/instrumentation/common/instrumentor.py
```diff
@@ -197,6 +197,14 @@ def _instrument(self, **kwargs):
                 if target_package == "agents.run" and target_method in ("run_single_turn", "_run_single_turn"):
                     logger.debug(f"method {target_method} not found in {target_package} (SDK version compatibility)")
                     continue
+                # openhands-sdk added the async API (arun/astep/_arun_safe) after 1.21
```
prasad-okahu: Can we just continue for all AttributeError cases? It's not a good idea to add meta model specific checks in the common/core. We want to keep the individual metamodel handling as isolate as possible.

prasad-okahu (changes requested):
Looks fine. Can you please add integration tests as well? Also, another feedback inline.

apptrace/src/monocle_apptrace/instrumentation/common/instrumentor.py
imohammedansari: Done in 3a04a63. Replaced the openhands-specific block with a generic `except AttributeError` clause: wrapt raises `AttributeError` when a listed method is absent from the installed SDK (a missing module still raises `ModuleNotFoundError`, handled separately), so that is a version-compatibility case for every framework and is now logged at debug uniformly. The pre-existing `agents.run` / `run_single_turn` special case is subsumed by it and removed, so no per-framework knowledge is left in common/core. Unit coverage added in `test_instrumentor_api.py::TestInstrumentorVersionCompatibility`.

PR #710, opened 2026-07-11
litellm: streaming span support and OpenAI Responses API coverage

imohammedansari:
## Proposed changes

Closes two litellm-layer gaps: **streaming inference spans carry no output/tokens** (the span closes before the stream is consumed), and **`litellm.responses()` produces no inference spans at all** (the Responses API goes through litellm's raw-httpx handler, which nothing wraps — this is the path gpt-5/o-series models take). Both were surfaced while instrumenting OpenHands, whose LLM layer is litellm, but they apply to every litellm application and are therefore a separate PR from the OpenHands metamodel (#708).

## Changes

1. **Streaming support on the litellm inference entity** (`metamodel/litellm/entities/inference.py`)
   - `is_auto_close` defers span close when `optional_params.stream` is set; `response_processor` reuses the existing OpenAI-format stream processor, so the span closes when the stream is exhausted, hydrated with the accumulated output, token usage, and finish_reason.
   - New `INFERENCE_STREAMING` variant (unconditional defer) wired to `OpenAIChatCompletion.async_streaming` — litellm's async streaming bypasses `acompletion` and routes here, and this method receives no `stream` flag in `optional_params` (litellm sets `data["stream"]` internally). Previously async streaming produced **no inference span at all**.

2. **Stream-processor fixes** (`metamodel/openai/openai_stream_processor.py`, `common/stream_processor.py`) — three real data-loss bugs in the shared OpenAI-format engine, each reproduced before fixing:
   - **Usage capture**: litellm's `CustomStreamWrapper` attaches usage to a chunk that still carries an (empty) delta, so `handle_chunk` returned True and `handle_completion` never saw it — tokens were silently lost. `handle_chunk` now also captures usage.
   - **Tool-argument accumulation**: a streamed tool call arrives as one id-bearing fragment followed by id-less argument deltas; `assemble_data` kept only the first fragment, so `tool_arguments` was always empty. Arguments are now accumulated per `tool_call.index`.
   - **Direct `__next__`/`__anext__` consumers**: the patcher only covered `__iter__`/`__aiter__`, but litellm pulls the inner stream via direct `__anext__()` attribute calls — spans never closed and were dropped. Both entry points are now patched, with an ownership flag so `async for` consumers don't double-process.

3. **OpenAI Responses API coverage** (`metamodel/litellm/entities/responses.py`, `_helper.py`, `methods.py`)
   - New `RESPONSES` entity wrapping `BaseLLMHTTPHandler.response_api_handler` / `async_response_api_handler` with input/output/usage/finish-reason accessors over `ResponsesAPIResponse` (function_call and message output items), plus streaming support via the same processor (SSE `response.*` events were already handled).

4. **Non-streaming tool-call output** (`_helper.py`): tool-call responses carry `content=None`, so `data.output` read `{"assistant": null}`; the helper now surfaces the tool calls (`{"assistant": [{tool_id, tool_name, tool_arguments}]}`), mirroring the openai metamodel helper. Applies to every litellm-based agent framework.

5. **Tests**: `tests/unit/test_litellm_streaming_responses.py` — 8 tests: fragment accumulation, delta-chunk usage capture, stream-result message extraction, `optional_params` stream detection, the `INFERENCE_STREAMING` variant, and the Responses accessors (input/output/finish_reason/usage). All stand-ins, no network.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

**Why the shared stream processor had to change:** the litellm entity reuses `OpenAIStreamProcessor` (as hugging_face already does), and the three fixes above are bugs for *any* OpenAI-format stream, not litellm specials — e.g. empty `tool_arguments` affects the openai metamodel's own streaming today. The alternative (a litellm-only subclass duplicating the processor) would fork shared logic to preserve known data loss.

**Behavior change to be aware of:** with streaming, the litellm inference span now closes when the stream is exhausted instead of at method return. A stream that is never consumed will not export its inference span (previously an empty-output span was exported); this matches the openai metamodel's existing streaming semantics.

**Validated with real runs** (`gpt-4o-mini`, `gpt-5-nano`): sync streaming (output + tokens + accumulated tool args), async streaming (previously span-less), and Responses API sync (previously span-less) — plus end-to-end through OpenHands. `tests/unit`: 506 passed; all failures present are byte-identical on clean `main` (pre-existing environment issues, verified by stash/re-run).

**Review round (adversarial review applied):** the async Responses path no longer leaks a never-ended span — `response_api_handler(_is_async=True)` returns an unawaited coroutine, so its sync span is now skipped via the existing `litellm_sync_handler` (extended to recognize `_is_async`), mirroring how chat-completion sync shims are already handled; close is idempotent (`fire_close_once` guards all six close sites — post-exhaustion re-iteration no longer re-fires span end/hydrate); generator early-exit releases stream ownership so a direct `next()` drain still processes and closes; index-less tool-argument fragments attach to the most recent tool call instead of being dropped, and id-only fragments no longer create ghost entries; streamed Responses SSE tool calls are assembled (`response.output_item.added` / `response.function_call_arguments.delta`) with correct `tool_calls` finish reason. 14 unit tests including direct-`next()` consumption, non-streaming tool-call extraction, close-once, and the `_is_async` skip. Live re-validated: async Responses emits exactly one closed span (no leaked twin); sync + async chat streaming carry output and tokens. Known scope limit: async streaming for non-OpenAI providers (Azure/Vertex/Bedrock/Anthropic dedicated `async_streaming` methods) is unchanged from main — wraps for those need provider-native chunk validation and are a follow-up.

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/common/stream_processor.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/common/stream_processor.py
@@ -71,6 +71,15 @@ class StreamState:
         """Mark stream as closed with current timestamp."""
         self.stream_closed_time = time.time_ns()
 
+    def fire_close_once(self) -> bool:
+        """Return True exactly once; guards span_processor against re-firing
+        when an exhausted stream is iterated again (span.end()/hydrate on an
+        already-ended span, duplicated assemble_data)."""
+        if getattr(self, "_close_fired", False):
+            return False
+        self._close_fired = True
+        return True
+
 
 class BaseStreamProcessor(ABC):
     """Base class for streaming processors using Template Method pattern.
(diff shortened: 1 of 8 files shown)


imohammedansari:
Superseded by #724, which consolidates this work on top of #641 (keeping its LiteLLMStreamProcessor and integration tests as the base) so the two streaming efforts land as one coherent change. Leaving this open for reference until #724 is reviewed; happy to close it whenever maintainers prefer.

imohammedansari:
Superseded by #726, which is trimmed to the minimal additive, non-streaming litellm changes (OpenAI Responses API coverage + non-streaming tool-call output) and touches only `metamodel/litellm/`. Streaming and the shared stream-engine changes remain in #641.

PR #712, opened 2026-07-12
Improve OTLP interoperability for GenAI traces

njpatel:
## Proposed changes

This change improves Monocle’s interoperability with OpenTelemetry-compatible observability backends without adding a vendor-specific exporter or dependency.

It:

- Documents authenticated OTLP/HTTP export using standard `OTEL_EXPORTER_OTLP_*` endpoint and header configuration.
- Adds optional OpenTelemetry `gen_ai.*` semantic attributes alongside Monocle’s existing metamodel attributes.
- Enables this enrichment automatically when the built-in OTLP exporter is configured, with explicit programmatic and environment-variable overrides.
- Maps Monocle spans to appropriate GenAI operations, including `chat`, `embeddings`, `retrieval`, `execute_tool`, and `invoke_agent`.
- Preserves existing `gen_ai.*` attributes rather than replacing instrumentation supplied by applications or other libraries.
- Fixes the non-isolated span path so `MONOCLE_ISOLATE_SPANS=false` produces correctly parented OpenTelemetry spans instead of yielding the span context manager.
- Adds regression tests and updates the user guide for the new behavior.

## Types of changes

## Checklist

- [ ] I have signed the CLA
- [x] I have added the necessary documentation
- [x] Any dependent changes have been merged and published in downstream modules — no downstream changes are required

## Testing

- 54 affected and adjacent unit tests pass.
- Python compilation and diff checks pass.
- Verified a real multi-span GenAI trace through an authenticated OTLP/HTTP endpoint.
- Verified that non-isolated Monocle spans retain the active OpenTelemetry parent and trace ID.
- Verified parsing of percent-encoded authorization and tenant headers.

The complete optional-integration test suite was not run locally because its Flask, Haystack, GCS, LangChain, PyTorch, and other optional dependencies were not installed.

## Further comments

The implementation deliberately keeps the export path generic. Monocle’s existing OTLP exporter already provides the required transport and supports standard OpenTelemetry authentication headers.

The additional compatibility layer only enriches exported spans with standard GenAI attributes. Monocle’s existing `entity.*`, workflow, event, and metadata attributes remain unchanged, preserving compatibility with current Monocle and Okahu consumers.

GenAI enrichment defaults to `auto`, meaning it is enabled when the built-in OTLP exporter is configured. Applications using custom span processors can explicitly enable or disable it through `otel_genai_semconv` or `MONOCLE_OTEL_GENAI_SEMCONV`.

### DIFF
--- a/apptrace/src/monocle_apptrace/exporters/monocle_exporters.py
+++ b/apptrace/src/monocle_apptrace/exporters/monocle_exporters.py
@@ -17,6 +17,10 @@ monocle_exporters: Dict[str, Any] = {
     "memory": {"module": "monocle_apptrace.exporters.base_exporter", "class": "MonocleInMemorySpanExporter"},
     "console": {"module": "opentelemetry.sdk.trace.export", "class": "ConsoleSpanExporter"},
     "otlp": {"module": "opentelemetry.exporter.otlp.proto.http.trace_exporter", "class": "OTLPSpanExporter"},
+    "otlp-genai-semconv": {
+        "module": "opentelemetry.exporter.otlp.proto.http.trace_exporter",
+        "class": "OTLPSpanExporter",
+    },
     "gcs" : {"module": "monocle_apptrace.exporters.gcp.gcs_exporter", "class": "GCSSpanExporter"},
     "postgres": {"module": "monocle_apptrace.exporters.postgres.postgres_exporter", "class": "PostgresSpanExporter"},
     "paygentic": {"module": "monocle_apptrace.exporters.paygentic.paygentic_exporter", "class": "PaygenticSpanExporter"}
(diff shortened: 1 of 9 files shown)


prasad-okahu:
@njpatel , thanks for the patch. Overall look fine to me. 
My only feedback is that we should create a new name (eg "otlp-genai-semconv") for OTLP exporter with these genai attribute mapping as default, and leave the behavior of the existing "otlp" exporter as is.

njpatel:
makes sense, pushed the changes!

njpatel:
@prasad-okahu i couldn't find the CLA to sign, do you have a link?

prasad-okahu:
> @prasad-okahu i couldn't find the CLA to sign, do you have a link?

We just need the DCO (signed commit) which is already there for this PR. We are good on that front.

prasad-okahu (approved):

prasad-okahu:
@njpatel , merged the PR. Thanks for your contribution!

PR #724, opened 2026-07-15
[claude-skill] litellm: async streaming, Responses API, stream-engine fixes (extends #641)

imohammedansari:
## Proposed changes

**Builds directly on #641** (@shambhavigoel-13's litellm streaming implementation — its commits and authorship are included in this branch unchanged, with its `LiteLLMStreamProcessor` and integration tests as the base). This PR layers the remaining litellm gaps on top, so the two land as one coherent story; it supersedes #710, which implemented an overlapping approach independently.

What #641 provides (unchanged): sync chat-completion streaming with deferred span close, chunk-text accumulation, tool-call assembly, multi-provider usage-key aliases, OpenAI + Azure integration tests.

What this PR adds on top:

1. **Async streaming** — litellm's async streaming bypasses `acompletion` and routes to `OpenAIChatCompletion.async_streaming`, which nothing wrapped: async streaming produced **no inference span at all**. A new `INFERENCE_STREAMING` entity variant (unconditionally deferred close — this method receives no `stream` flag in `optional_params`) wires it up, reusing #641's processor.
2. **Stream-engine fixes** (`common/stream_processor.py`, inherited by every `BaseStreamProcessor` subclass including #641's): consumers that pull via direct `__next__()`/`__anext__()` attribute calls (exactly what litellm's `CustomStreamWrapper` does with the inner stream) bypassed the `__iter__`/`__aiter__`-only patching — spans never closed and were silently dropped; both entry points are now patched with an ownership flag so `for`/`async for` consumers don't double-process. Close is idempotent (`fire_close_once`, guards all six close sites), and generator early-exit releases ownership so a direct drain still completes the span.
3. **OpenAI Responses API** — `litellm.responses()`/`aresponses()` go through `BaseLLMHTTPHandler.response_api_handler`/`async_response_api_handler` (raw httpx; the path gpt-5/o-series models take) and produced **zero inference spans**. New `RESPONSES` entity with input/output/usage/finish-reason accessors, streamed SSE tool-call assembly (`response.output_item.added` / `response.function_call_arguments.delta` in the OpenAI-format processor), and the sync shim is skipped when `_is_async=True` via the existing `litellm_sync_handler` (the sync call returns an unawaited coroutine — spanning it leaked a never-ended span).
4. **`LiteLLMStreamProcessor` now derives from `OpenAIStreamProcessor`** (the same pattern `HFStreamProcessor` already uses): chunk/SSE handling and tool-call assembly are inherited, removing ~70 duplicated lines, while the LiteLLM-specific overrides are kept — multi-provider usage-key normalization (`candidates_token_count` etc.) and same-chunk text+finish_reason processing. Behavior verified unchanged via the PR's own integration tests.
5. **Non-streaming tool-call output** — tool-call responses carry `content=None`, so `data.output` read `{"assistant": null}`; the helper now surfaces the tool calls (mirrors the openai metamodel helper).

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

**Validation (all real runs, `gpt-4o-mini` / `gpt-5-nano`, on this exact branch):** five-path probe — sync streaming with tool calls (#641's processor + assembly), non-streaming tool call, sync Responses, async streaming (previously span-less), async Responses (exactly one closed span — no leaked twin) — every span carries output and token usage. #641's own integration test `test_litellm_openai_stream.py` passes on the merged branch. Unit suite: 530 passed, 14 new unit tests (`test_litellm_streaming_responses.py`: direct-`next()` consumption, close-once, fragment accumulation, `_is_async` skip, Responses accessors, non-streaming tool-call extraction); the remaining failures are byte-identical on clean `main` (pre-existing environment issues).

**Relationship to other PRs:** includes #641 (please review that one first or together — its commits are this branch's base); supersedes #710 (same goals, pre-consolidation). The OpenHands framework metamodel (#708) is independent and unaffected.

**Known scope limit:** async streaming for non-OpenAI providers (Azure/Vertex/Bedrock/Anthropic dedicated `async_streaming` methods) is unchanged from main — wraps for those need provider-native chunk validation and are a follow-up.


imohammedansari:
Superseded by #726, which is trimmed to the minimal additive, non-streaming litellm changes (OpenAI Responses API coverage + non-streaming tool-call output) and touches only `metamodel/litellm/`. Streaming and the shared stream-engine changes remain in #641.

PR #721, opened 2026-07-14
Implemented generic API that verifies any attribute/entity of span.

shambhavigoel-13:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [CONTRIBUTING](https://github.com/monocle2ai/monocle/blob/main/CONTRIBUTING.md) doc
- [ ] I have signed the CLA
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #730, opened 2026-07-17
Add CSV → fluent test-case adapter to monocle_test_tools

cjames-okahu:
## Proposed changes

Adds a **CSV → fluent eval test-case adapter** to `monocle_test_tools` so evaluation test cases can be curated in a flat spreadsheet and run as parametrized pytest tests through the fluent `TraceAssertion` path.

**Scope (v0): evaluation tests only.** Following review, the adapter is deliberately narrow — it does **not** mirror the open fluent assertion set. Each row asserts an eval label and, optionally, two operational guard rails on the run. The loader bridges exactly **three stable, eval-relevant fluent methods**:

| CSV concern | Fluent method | Role |
| --- | --- | --- |
| `expected` / `not_expected` | `check_eval` | the label — the tuning target |
| `max_tokens` | `under_token_limit` | **budget** guard rail |
| `max_duration_ms` | `under_duration(units="ms")` | **effort** guard rail |

This keeps the adapter immune to fluent API churn (a new assertion never touches the CSV schema) and stops it competing with the JSON `TestCase` executor — it is purpose-scoped to evals, which the `TestCase`/`validate()` path does not run.

New public surface (in `monocle_test_tools`):
- `load_cases_from_csv(path) -> list[CsvCase]` — parse + validate a CSV (usable standalone)
- `CsvCase.run(asserter, **check_eval_kwargs)` — maps one row onto the fluent chain
- `@monocle_csv_cases("cases.csv")` — parametrizes a one-line test stub over the rows

**Design principle — config in code, data in CSV.** The test stub owns everything constant across the sheet (the evaluator via `with_evaluation(...)`, the eval `template_path`, and the trace source — fixed to `okahu` in this version); the CSV owns only what varies per row. Example stub:

```python
@monocle_csv_cases("cases.csv")
def test_cases(monocle_trace_asserter, case):
    case.run(monocle_trace_asserter.with_evaluation("okahu"), template_path=TEMPLATE_PATH)
```

Columns: `case_id, fact_id, workflow_name, fact_name, expected, not_expected, max_tokens, max_duration_ms, notes`.
- `expected`/`not_expected` accept multi-values (pipe-delimited or JSON array).
- `fact_id` identifies the evaluated fact — it equals the trace id when `fact_name=traces` and is a finer-grained id otherwise.
- A row must declare an `expected`/`not_expected` label; a guard rail alone is not an eval test.

**Notable design points**
- Targets the fluent `TraceAssertion` path (not `TestCase`): the `TestCase`/`validate()` path does not execute evals (`_evaluate_span` is unused) and `TestCase` has no trace source, so the fluent path is the only one that runs evals *and* replays a trace.
- Honors the PR #721 collected-assertions model: each row's independent assertions are invoked on the base asserter (not threaded), and failures surface through the existing pytest plugin.
- Fact-name validation is delegated to the Okahu evaluator (no hardcoded fact list in the adapter).
- Load-time validation with clear `line N: …` messages, including duplicate-`case_id` and required-column checks.
- Stdlib only — **no new runtime dependencies**; Python ≥3.8.

### Out of scope for v0 (future considerations)

- **General non-eval assertions from CSV** (tool/agent calls, input/output, arbitrary fluent methods) and **multi-condition rows** (e.g. asserting two different tools). Deferred; if pursued, the intended shape is a **column-agnostic raw loader + code-side mapping** — plus a `SUPPORTED_ASSERTIONS` API exposed from the fluent module — rather than a growing column→method schema. This is what avoids the backward-compat churn a full mirror would incur.
- **Non-okahu trace sources** (`local`/`file`) — v0 fixes the source to `okahu`.
- **Sub-trace fact selectors / session-scope location**, and wiring `TestSpan.eval` + trace source into `TestCase`.
- **Result-matrix capture for non-eval assertions.**

### Design note: CSV parsing uses the stdlib `csv` module (not pandas)

Recording the rationale, since reviewers may wonder why this doesn't use pandas.

**Decision:** parse with the standard-library `csv` module; **do not** add pandas (or any third-party CSV lib) as a dependency.

**Why:**
- **Dependency weight.** pandas pulls in numpy and tens of MB. `monocle_test_tools` is a pip-installed test library, and this PR deliberately adds **no new runtime dependencies**.
- **It wouldn't simplify the code.** The schema is small and flat, and the cell semantics are custom — multi-value `expected`/`not_expected` (pipe- or JSON-delimited) and friendly `path:line` error messages. None of that comes from pandas.
- **Type inference actively fights us.** pandas coerces empty cells to `NaN` and guesses dtypes, so we'd read with `dtype=str, keep_default_na=False` and re-parse every cell ourselves anyway.
- **Author-facing errors need source line numbers.** `csv.DictReader` + `enumerate` yields the true CSV line for each row, which is what non-engineer curators need. A DataFrame gives a row *index*, not the file line.

**Format-swappability is preserved.** The reader is isolated as `read_rows(path) -> list[dict]`, so a future Excel/Google-Sheets reader can be added **behind that seam without touching the mapper or `CsvCase`**. If/when that lands it would be an **optional** extra (e.g. `monocle_test_tools[excel]`), never a hard dependency.

Tracking issue: okahu/monocle_backlog_tracking#215.

## Type of change

## Checklist

- [x] I have signed the CLA

## Further comments

Additive only — existing suites, the fluent API, and the pytest plugin are untouched. New unit tests cover the reader, cell parsing, row mapping/validation (including the no-label rejection and guard-rail mapping), `CsvCase.run` (via a recording double), and the `monocle_csv_cases` decorator; a copy-ready example (`test_tools/examples/`) includes an offline load smoke test; and an opt-in `integration`-marked test replays an okahu trace end-to-end. Full `test_tools/tests/unit` suite: 212 passed locally (the one pre-existing collection error is an unrelated optional `google.adk` dep, not installed).

All commits are DCO `Signed-off-by`.

### DIFF
--- a/test_tools/src/monocle_test_tools/csv_cases.py
+++ b/test_tools/src/monocle_test_tools/csv_cases.py
@@ -1,15 +1,24 @@
-"""CSV -> fluent test-case adapter for monocle_test_tools.
+"""CSV -> fluent eval test-case adapter for monocle_test_tools.
 
-Loads a flat CSV of okahu-trace test cases and drives each row through the
-existing fluent TraceAssertion API via ``@monocle_csv_cases`` + ``CsvCase.run``,
-so test data can be curated in a spreadsheet.
+Loads a flat CSV of okahu-trace *evaluation* test cases and drives each row
+through the existing fluent TraceAssertion API via ``@monocle_csv_cases`` +
+``CsvCase.run``, so eval test data can be curated in a spreadsheet.
+
+Scope (v0): evaluation tests only. Each row asserts an eval label
+(``expected``/``not_expected`` via ``check_eval``) and, optionally, operational
+guard rails on the run: a token budget (``max_tokens`` -> ``under_token_limit``)
+and an effort ceiling (``max_duration_ms`` -> ``under_duration``). The loader
+bridges exactly these three stable, eval-relevant fluent methods -- not the
+open set of assertions. General non-eval assertions (tool/agent calls,
+input/output, arbitrary fluent methods) and multi-condition rows are
+intentionally out of scope for v0; see the PR for the rationale and future
+considerations.
 
 Config-in-code, data-in-CSV: the test stub owns everything constant across the
 sheet (the evaluator via ``with_evaluation(...)``, the eval ``template_path``,
 and the trace source, fixed to ``okahu`` in this version); the CSV owns only
-what varies per row (trace id, workflow, expected labels, and per-row
-assertions). One row = one test; a row may carry several assertions, each run
-independently against the same loaded trace.
+what varies per row (fact id, workflow, expected labels, guard rails).
+One row = one test.
 
 Example test stub::
 
@@ -22,26 +31,12 @@ import csv
 import inspect
 import json
 import os
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Dict, List, Optional, Union
 
 import pytest
 
-from monocle_test_tools.fluent_api import TraceAssertion
-
-REQUIRED_COLUMNS = ("case_id", "id", "workflow_name")
-
-FACT_NAME_VALUES = (
-    "traces", "inferences", "agentic_turns", "agentic_sessions",
-    "agent_invocation", "tool_execution", "commits", "conversations",
-    "test_runs", "tests",
-)
-
-# Columns that constitute an assertion; a valid row has >= 1 populated.
-ASSERTION_COLUMNS = (
-    "expected", "not_expected", "called_tool", "called_agent",
-    "token_limit", "duration_ms", "has_input", "has_output", "extra_json",
-)
+REQUIRED_COLUMNS = ("case_id", "fact_id", "workflow_name")
 
 
 class CsvCaseError(ValueError):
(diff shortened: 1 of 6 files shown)


cjames-okahu:
A few follow-up observations from adopting this adapter in the eval-tune-kit (`okahu/skills`). Landing it on the fluent-API path was the right call — the `TestCase.validate()` path doesn't execute evals, so this is the correct layer.

**1. `FACT_NAME_VALUES` will drift from `okahu_eval._map_fact_name`**

`csv_cases.py` defines `FACT_NAME_VALUES` as a hardcoded 10-value tuple (`traces`, `inferences`, `agentic_turns`, `agentic_sessions`, `agent_invocation`, `tool_execution`, `commits`, `conversations`, `test_runs`, `tests`). This is a hand-copy of the supported levels in `evals/okahu_eval.py`'s `_map_fact_name`. When `_map_fact_name` gains a new level (or renames one), `FACT_NAME_VALUES` will silently lag, causing the adapter to reject a valid `fact_name` at parse time before the eval module ever sees it.

Preferred approach: derive the valid set from `_map_fact_name` directly (pull its branch keys at import time, or expose a `SUPPORTED_FACT_NAMES` constant from `okahu_eval.py` and reference it from `csv_cases.py`). That keeps the two in sync automatically.

**2. Consider renaming the `id` column to `fact_id`**

The current column name `id` is spreadsheet-ambiguous — Excel auto-generates an `id` column, and the name gives no hint that it holds a trace/fact identifier. More importantly, the adapter supports `fact_name` values below trace grain (e.g. `tool_execution`, `agentic_turns`), where one trace yields many facts and the value in this column identifies a specific sub-trace fact rather than the trace itself. `trace_id` would be an improvement over `id`, but it's still misleading for sub-trace rows.

The general term is `fact_id` — the identifier of the evaluated fact, which equals the trace id when `fact_name=traces` but is a finer-grained id otherwise. Renaming now, before the adapter reaches a stable release, avoids a breaking rename later. Suggested header:

```
case_id,fact_id,workflow_name,fact_name,expected,not_expected,notes
```

**3. The eval-tune-kit now adopts this adapter**

For reference: the eval-tune-kit (`okahu/skills`, `kits/eval-tune-kit/`, PR okahu/skills#14) has adopted `@monocle_csv_cases` and deleted its bespoke `fact_loader.py`. The benchmark test is now a thin adapter stub; the kit keeps only the N×-replay runner and cross-run stability analysis.

One follow-up that came out of this migration: the eval runner (`check_eval`) should validate `expected`/`not_expected` against the template's declared enums before submitting any eval jobs — an out-of-enum label today silently produces a permanent false mismatch. The kit's retired `fact_loader.py` had this guard; moving it into `check_eval` (which already has `template_path`) makes it universal. Filed as a backlog issue: okahu/monocle_backlog_tracking#216.

test_tools/src/monocle_test_tools/csv_cases.py
```diff
@@ -0,0 +1,294 @@
...
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Union
+
+import pytest
+
+from monocle_test_tools.fluent_api import TraceAssertion
+
+REQUIRED_COLUMNS = ("case_id", "id", "workflow_name")
+
+FACT_NAME_VALUES = (
```
prasad-okahu: We shouldn't hardcode the Okahu fact names here. I suggest we remove this check and let it be handled in the Okahu evaluator if it's used with unsupported fact to evaluate.

test_tools/src/monocle_test_tools/csv_cases.py
```diff
@@ -0,0 +1,294 @@
...
+    "traces", "inferences", "agentic_turns", "agentic_sessions",
+    "agent_invocation", "tool_execution", "commits", "conversations",
+    "test_runs", "tests",
+)
+
+# Columns that constitute an assertion; a valid row has >= 1 populated.
+ASSERTION_COLUMNS = (
```
prasad-okahu: Same with the assertions. I suggest that we make a generic error message here. A list of supported assertions should be provided as an API from fluent module to facilitate this. I guess it can be done as a follow up ticket.

prasad-okahu (commented):
@cjames-okahu, I see a few issues with the approach
- You can't have multiple condition checks eg I want to test a two different tools called
- When we add a new assertion in fluent, it's going be very messy to support it without breaking backward compatibility 
- Note that there's already a json based test framework executor in monocle test which handles this more generically ...

My suggestion is that we leave the CSV structure ie the columns and values to test developers. The monocle framework will make it easier to load it from CSV for parameters test, rather that facilitating full test instructions via CSV. For example,

test_tools/src/monocle_test_tools/csv_cases.py
cjames-okahu: Removed the hardcoded `FACT_NAME_VALUES` tuple and the parse-time check. Fact-name validation is now delegated to the Okahu evaluator, so the adapter no longer duplicates (and can no longer drift from) the supported-levels list.

test_tools/src/monocle_test_tools/csv_cases.py
cjames-okahu: Moot for v0: the reflection path (`extra_json` + `_validate_extra_steps`) is removed entirely with the eval-only scope, so there is no assertion-name validation to make generic here. The suggestion to expose a `SUPPORTED_ASSERTIONS` API from the fluent module is captured as a future item for the general (non-eval) CSV adapter, and is noted in the PR description under Out of scope.

cjames-okahu:
Reshaped to a v0 scope of **evaluation tests only**, which addresses the three concerns directly:

- **Multiple condition checks / new fluent assertions.** The adapter no longer mirrors the fluent API. It bridges exactly three stable, eval-relevant methods — `check_eval` (the label being tuned) plus two operational guard rails, `under_token_limit` (`max_tokens`) and `under_duration` (`max_duration_ms`). A new fluent assertion never touches the CSV schema, so there is no backward-compat surface to maintain. Multi-condition and general non-eval assertions are explicitly out of scope for v0.
- **Existing JSON executor.** Agreed we should not build a second generic test executor. This is purpose-scoped to evals, which the JSON `TestCase`/`validate()` path does not run — so it complements rather than duplicates it.
- **Leaving CSV structure to test developers.** Noted for the general case: if we later support non-eval assertions from CSV, the intended shape is a column-agnostic raw loader + code-side mapping (plus a `SUPPORTED_ASSERTIONS` API from fluent), not a growing column→method schema. Captured under Out of scope in the PR description.

Inline comments addressed in-thread. PR description updated with the v0 scope and the out-of-scope/future list.

prasad-okahu (approved):

PR #731, opened 2026-07-17
feat(test-tools): filtered eval flow (check_eval_filtered) for built-in + custom templates

cjames-okahu:
## Summary

Adds a first-class **Filtered** eval flow to `monocle_test_tools`:
`with_filtered_source(...).check_eval_filtered(...)`. It submits **one async
`/v1/eval/jobs` job** scoped by a filter (workflow + time window + fact level; **no
`trace_id`**), lets the server discover every matching fact, reconciles each against a
**single blanket** expectation, and reports per fact — with **built-in and custom
template** support (the exact `eval_name` XOR `template_path`/`template` switch that
`check_eval` uses).

Implements **Phases 1–5** of the plan. Server follow-ups (Phase 7) and the
`evaluation-testing` adoption/retirement (Phase 6) are out of scope for this PR.

- **Design:** okahu/monocle_backlog_tracking `designs/2026-07-09-monocle-batch-custom-eval-support.md`
- **Plan:** okahu/monocle_backlog_tracking `plans/2026-07-09-monocle-batch-custom-eval-implementation.md`
- **Tracking issue:** okahu/monocle_backlog_tracking#211
- **Stacked on:** #722 (the shared `eval_matrix` recorder). This PR targets `main`
  but depends on #722; **do not merge until #722 lands**. Until then the diff below
  also includes #722's commits (the `eval_matrix.py` recorder it extends) — it
  auto-narrows to just this PR's changes once #722 merges to `main`.

## What's added (all additive)

- **`evals/okahu_filtered_eval.py`** (new) — pure helpers (`normalize_fact_id`,
  `has_label`, `grade_label`, `build_filtered_report`, `labeled_fact_ids`) plus the
  `OkahuFilteredEval` client (`submit`, `poll_status`, `get_job_result_fact_ids`,
  `query_results` custom-aware, `poll_for_coverage`, `run_filtered`). Uses **only
  `requests` + stdlib** — deliberately independent of the bert-score/transformers
  evaluators.
- **`fluent_api.py`** — `with_filtered_source(...)`, `check_eval_filtered(...)`, and
  report accessors (`get_filtered_eval_report` / `get_filtered_eval_failures` /
  `write_filtered_eval_report`).
- **`eval_matrix.py`** — widened the row schema with `fact_id`/`workflow`/`job_id`
  (empty-string defaults for interactive rows) and added
  `record_eval_rows_from_report`, so the filtered flow feeds the **same** accumulator
  and `pytest_sessionfinish` JSON writer as the interactive path — no duplicate matrix
  code.

## Two load-bearing invariants (or filtered assertions are silently wrong)

1. **Bare-hex `fact_id`** — every submit/query/reconcile strips a leading `0x`
   (eval_results rows are keyed bare-hex).
2. **`job_id` gate** — a fact counts as covered only when the returned row's `job_id`
   equals the just-submitted job, so an assertion never settles on a prior run's label.

## Coverage / vacuous-run guard

`min_facts` (default 1) fails a run that discovered too few facts — filtered mode has no
fixed denominator, so this prevents a false pass on zero data.

## Tests

- **Unit (no network, `unittest.mock`):** `test_okahu_filtered_eval.py` (helpers +
  client + `run_filtered`), `test_check_eval_filtered.py` (fluent validation +
  end-to-end wiring with the client mocked), and `test_eval_matrix.py` (widened schema +
  `record_eval_rows_from_report`). **225 unit tests pass** (the one collection error in
  `test_non_llm_core_examples.py` is the pre-existing missing `google.adk` dev dep,
  unrelated to this PR).
- **Integration (opt-in, skipped without creds):**
  `tests/integration/test_filtered_custom_eval_live.py` — one live filtered + custom
  template smoke test.

## Verify-at-execution note

`OkahuFilteredEval.get_job_result_fact_ids` reads the fact-id from each job-detail
`results` row; the exact field name (`fact_id` vs `id` vs nested) must be **confirmed
against a live job** (plan Phase 5 Step 2) — the live integration run is where that is
checked.


test_tools/src/monocle_test_tools/evals/okahu_filtered_eval.py
```diff
@@ -0,0 +1,287 @@
...
+def has_label(result) -> bool:
+    """True if an API result row carries a usable eval label."""
+    return bool(
+        result
+        and result.get("eval_found")
+        and (result.get("eval_result") or {}).get("label")
+    )
+
+
+_COUNT_KEY = {"pass": "passed", "drift": "drift", "error": "errors"}
```
prasad-okahu: Can we change the "drift" to "fail"? That's testing convention.

test_tools/src/monocle_test_tools/evals/okahu_filtered_eval.py
```diff
@@ -0,0 +1,287 @@
...
+        Exactly one of `eval_name` (built-in template) or `template` (inline custom).
+        """
+        if bool(eval_name) == bool(template):
+            raise ValueError("Provide exactly one of 'eval_name' or 'template'.")
+        url = f"{self.eval_base}/v1/eval/jobs"
+        params = {
+            "workflow_name": workflow_names,
+            "start_time": start_time, "end_time": end_time,
+            "fact_name": fact_name, "breakdown_filter": fact_name,
+            "shadow_eval": "true",
```
prasad-okahu: I don't think we should hardcode shadow_eval to true. The eval job should run against the existing traces already loaded in Okahu. The shadow eval only suppose to look at the temporary traces generated by the test itself.

test_tools/src/monocle_test_tools/fluent_api.py
```diff
@@ -813,6 +771,98 @@ def check_eval(self, eval_name:Optional[str] = None, expected:Optional[Union[str
 
         return self
 
+    def with_filtered_source(self, source: str = "okahu", *, workflow_name,
```
prasad-okahu: I think we should make the time range as an optional parameter for with_trace_source() rather than adding new APIs. For Okahu source it will accept either the id or time box (error if both are provided).
That would also eliminate the need for separate check_eval_filtered(). If the scope is set then use the filtered check else id based

test_tools/src/monocle_test_tools/evals/okahu_filtered_eval.py
```diff
@@ -0,0 +1,287 @@
...
+        url = f"{self.api_base}/v1/workflows/{workflow_name}/evals/query"
+        eval_names = ["custom"] if custom_name else [eval_name]
+        body = {
+            "eval_names": eval_names, "fact_name": fact_name,
+            "fact_ids": [normalize_fact_id(f) for f in fact_ids],
+            "start_time": start_time, "end_time": end_time,
+        }
+        resp = requests.post(url, headers=self.headers, json=body, timeout=60)
+        resp.raise_for_status()
+        results = resp.json().get("results", [])
```
prasad-okahu: - We should pin down a max size of fact IDs that this will support. With time box, you could have very large dataset that could cause token budget issues.
- The query API returns paginated results. The current implementation is not taking that into account.

I suggest that as a first stage, we set the max count to be 100 (or some smaller number). Set that as the page size as first iteration. If it gets over the pagesize, then raise error.
I don't think the Okahu eval submit API sets the max factID count. Once that's supported, we should use it there as well.

cjames-okahu:
## Reshaped the eval API per review (#3 single API, #4 pagination)

@prasad-okahu — heads-up that this PR now looks substantially different from the first review. Based on the discussion, we took the unification further than a cosmetic change, because two near-term capabilities (a large trace-**id list** routed to the async batch, and additional fact **attributes** as filters) turn "scope" into an open-ended predicate set that a fixed set of named methods can't model cleanly. Better to reshape the surface now than ship an API we already know will change.

**Design + plan (monocle_backlog_tracking):**
- `designs/2026-07-20-monocle-unified-okahu-eval-source-design.md` (revises the API-surface decision in the 2026-07-09 design)
- `plans/2026-07-20-monocle-unified-okahu-eval-source-implementation.md`

### What changed

**One source, one check (was two of each).** The time-window filter now lives on `with_trace_source`, and `check_eval` branches on scope — `with_filtered_source`/`check_eval_filtered` are gone.

```python
# Direct (unchanged behavior): id -> sync, spans fetched, structural asserts available
asserter.with_trace_source("okahu", id=tid, workflow_name="app") \
        .check_eval(eval_name="hallucination", expected=exp)

# Filtered (new): time window -> async job, eval-only, blanket expectation over discovered facts
asserter.with_trace_source("okahu", workflow_name="app", start_time=s, end_time=e) \
        .check_eval(eval_name="hallucination", expected="no_hallucination", min_facts=5)
```

- `id` XOR time window (error if both). Delivery (sync ≤10-id vs async job) is chosen by the selector and hidden — matching how the server already keys mode on `trace_id` presence, and the design's "delivery is implementation detail" stance.
- Uniform result: `get_eval_report()` / `get_eval_failures()` return the same shape in both modes (span mode = a 1-fact report), populated on pass **and** failure.
- Filter-only params (`min_facts`, `fail_threshold`, `max_facts`) raise a clear error if used in span mode.

**Comment #4 — pagination + size, located correctly.** The results query now paginates fully (`query_results` and `get_job_result_fact_ids` previously read only the first page — a silent under-count → gate passing on partial data). On the size cap: token spend happens at **job submit** (the server evaluates every discovered fact), so a low cap on the *result read* hides facts without saving tokens — and the design's own grounded run discovered ~97 facts, so a cap of 100 would break real gates. Instead we added a generous, env-configurable `max_facts` **guard** (`OKAHU_MAX_FACTS`, default 1000) that fails loudly on a runaway window. True budget control belongs in the server's `max_facts_per_job` at submit (deferred).

**Comment #1/#2 (from the earlier round):** grade vocab is `pass`/`fail`/`missing`; `shadow_eval=False` for filtered jobs (they grade traces already in Okahu, not shadow traces).

### Deferred (reserved seams, blocked on server work)
- **Large id-list → async batch** — the submit API has no `fact_ids` allowlist param yet; the scope dict reserves `ids`.
- **Attribute filters** — map to the server's `eval_filters` (deferred/shape-unverified); scope dict reserves `attributes`.

### Backward compatibility
Interactive `check_eval` and `with_trace_source(local/file/okahu+id)` are unchanged — the eval-tune-kit benchmark, evaluation-testing, and examples are untouched. Selector widened to a three-way XOR (`eval_name`/`template_path`/`template`); existing `eval_name`-only and `template_path`-only calls behave identically.

### Tests
244 unit tests pass; the live integration test was migrated to the unified API (collects + skips without creds).

### One open verification (needs live creds)
The `/evals/query` **pagination transport** (assumed `limit`/`offset` + short-page stop) and the job-detail **fact-id field name** are still unverified against a live job — flagged as verify-at-execution in the design. The unit tests mock HTTP, so they can't catch a wrong transport shape. Will confirm against a live run before we rely on filtered gates.

prasad-okahu (approved):

PR #733, opened 2026-07-17
 ADK: bind turn scope to native invocation_id

imohammedansari:
Bind the ADK turn scope (`scope.agentic.turn`) to ADK's native `invocation_id`, instead of a random, ADK-unrelated value — so a turn is identifiable end-to-end (telemetry ↔ ADK ↔ app) and the id lands on **every** span of the turn, including the workflow/root span.

This works on the ADK version Kahu runs today; it reads ADK's own generated `invocation_id`, so it does not depend on any specific ADK version or on the caller supplying an id.

### Core (`common/wrapper.py`)
`get_builtin_scope_names` gains a small opt-out flag (`skip_builtin_scope`): an entity marks that its builtin scope's value is set elsewhere (by its span handler), so the core does not auto-generate the random default. Backward-compatible — entities that don't set it behave exactly as before.

### ADK metamodel
- The `REQUEST` (turn) entity sets `skip_builtin_scope: True`, so the core does **not** mint a random turn scope for it.
- `AdkSpanHandler` sets `scope.agentic.turn = ctx.invocation_id` at the **root-agent boundary** (the earliest point ADK's id exists, before any LLM/tool child span) so the agent subtree inherits it, and stamps it onto the already-open turn/REQUEST span. `post_task_processing` copies it onto the workflow/root span (which opened before the id existed and doesn't re-hydrate scopes at close).

### Why the handler is needed
ADK mints `invocation_id` **inside** `run_async`, after the turn span and workflow span have already opened — so the value can't be set purely up-front. The handler captures it at the first point it exists and stamps the two ancestor spans. This is version-independent: it reads whatever id ADK generates.

## Types of changes
- [ ] Breaking change
- [ ] Documentation Update

## Checklist
- [ ] I have signed the CLA

## Further comments

**Coverage:** every span in the turn carries `scope.agentic.turn` — the REQUEST/turn span and the agentic subtree (agent, LLM, tools) via baggage, plus the workflow/root span via child→parent propagation in `post_task_processing` (reusing the existing `propogate_inference_info_to_parent_span` idiom).

**Backward compatibility:** unchanged behavior for other frameworks and for ADK apps generally — the turn id is ADK's own `invocation_id`; nothing must be passed by the caller.

**Verified end-to-end** on a real multi-agent ADK app, multi-session/multi-turn, on both `Runner.run_async` and the threaded sync `Runner.run`: turn id = ADK's `invocation_id`, unique per turn, consistent across the subtree, present on the workflow span.

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/common/wrapper.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/common/wrapper.py
@@ -767,8 +767,9 @@ def start_as_monocle_span(tracer: Tracer, name: str, auto_close_span: bool,
 def get_builtin_scope_names(to_wrap, instance=None, args=None, kwargs=None) -> tuple:
     """Return the builtin scope to open, as a tuple splatted into monocle_trace_scope():
        (name,)         -> auto-generate a value (default);
-       (name, value)   -> use this value (entity opts in via use_builtin_scope + scope_value);
-       (None,)         -> open no scope (value is set elsewhere, e.g. by the metamodel handler).
+       (name, value)   -> entity opted in via use_builtin_scope + scope_value; use this value
+                          (value None -> monocle_trace_scope auto-generates, i.e. graceful fallback);
+       (None,)         -> open no scope (non-agentic span).
     """
     output_processor = None
     if "output_processor" in to_wrap:
@@ -782,8 +783,8 @@ def get_builtin_scope_names(to_wrap, instance=None, args=None, kwargs=None) -> t
     span_type = output_processor.get("type", None) if isinstance(output_processor, dict) else None
 
     # An entity can supply the builtin scope's value itself (instead of the random default)
-    # via a `scope_value` accessor evaluated with the call arguments. If it yields no value,
-    # the scope is left unset (the value is provided elsewhere) rather than auto-generated.
+    # via a `scope_value` accessor evaluated with the call arguments. A None result falls
+    # back to an auto-generated value, so callers that don't supply one still get a scope.
     if isinstance(output_processor, dict) and output_processor.get("use_builtin_scope", False):
         scope_name = output_processor.get("scope_name", span_type)
         accessor = output_processor.get("scope_value")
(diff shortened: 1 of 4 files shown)


prasad-okahu (commented):
I prefer the common logic in the core and avoid more exception handling code in each meta model. How about we have a flag like use_builtin_scope and if it's set then use that value in setting scope with `monocle_trace_scope()`. That we we don't have to code all the span formatting in the metamodel.

```python
def get_builtin_scope_names(to_wrap) -> str:
...
  if isinstance(output_processor, dict) and output_processor.get("use_builtin_scope", False):
    return span_type, get_scope(span_type)
  else:
    return span_type
```
Then in all places replace  `monocle_trace_scope(get_builtin_scope_names(to_wrap))` with`monocle_trace_scope(*get_builtin_scope_names(to_wrap))` 

This way we don't need all the span reformatting in each metamodel that needs this solution.

imohammedansari:
> I prefer the common logic in the core and avoid more exception handling code in each meta model. How about we have a flag like use_builtin_scope and if it's set then use that value in setting scope with `monocle_trace_scope()`. That we we don't have to code all the span formatting in the metamodel.
> 
> ```python
> def get_builtin_scope_names(to_wrap) -> str:
> ...
>   if isinstance(output_processor, dict) and output_processor.get("use_builtin_scope", False):
>     return span_type, get_scope(span_type)
>   else:
>     return span_type
> ```
> 
> Then in all places replace `monocle_trace_scope(get_builtin_scope_names(to_wrap))` with`monocle_trace_scope(*get_builtin_scope_names(to_wrap))`
> 
> This way we don't need all the span reformatting in each metamodel that needs this solution.

Hey Prasad,

I looked into making this a common mechanism, similar to how we handle session_id, but it doesn't seem feasible for invocation_id right now.

The latest ADK (2.5.0), released recently, actually uses an invocation_id we pass into run_async. Older versions, including the one we're on, ignore it: run_async accepts the argument but generates its own id internally regardless, so we can't set our own.

So for now I kept this ADK-specific. The ADK handler reads the generated `invocation_id` and sets it on the scope. Once we upgrade to a newer ADK version and after it's more stable, we can revisit moving this into the common logic.

PR #741, opened 2026-07-22
fix: address issue #654 - Add Generic Template Support for Okahu Evals

### LINKED ISSUE #654: Add Generic Template Support for Okahu Evals
Current evaluations don't support generic templates as a fallback. They only support templates of specified types. 
The template verification method should also use generic template support as fallback in any template query.

Currently, template verification only matches templates where both name and group_by exactly match the requested (eval_name, fact_name). If no exact match exists, the evaluation fails. This occurs even when a generic template (group_by='generic') exists for that eval name and could support the evaluation request.

This is a problem for newer fact types like tool_execution, agent_invocation, and commits, which may only have generic templates on the backend. For example, check_eval("sentiment", fact_name="tool_execution") fails at template verification even though the backend can evaluate it using the generic sentiment template.

Proposed Change:

When no exact (eval_name, fact_name) template match is found, fall back to a template where name matches and group_by == 'generic'. If a generic fallback is used, log an info message. If neither an exact match nor a generic fallback exists, raise the existing error.

Files:
- test_tools/src/monocle_test_tools/evals/okahu_eval.py

kunalrawat425:

Closes #654


kunalrawat425:

PR #743, opened 2026-07-22
fix: address issue #634 - [Feature Request]: Custom instrumentation decorator variations — span-only and custom attributes

### LINKED ISSUE #634: [Feature Request]: Custom instrumentation decorator variations — span-only and custom attributes
Current behavior: The monocle_trace_method decorator creates a custom span and captures all input (args/kwargs) and output (return value) as two events. There is no option to create a span without events, or to add custom attributes to the span.

Expected behavior: Provide decorator variations that support:
1. Span-only mode — creates the span but does not capture input/output events (useful when the data is too large or sensitive to record)
2. Custom attributes — allows the user to attach arbitrary key-value attributes to the span

Where to start: The existing decorator is in apptrace/src/monocle_apptrace/instrumentation/common/method_wrappers.py at line 171. The CUSTOM_SPAN_PROCESSOR output processor controls event creation — a flag or parameter to skip it would handle the span-only case. Custom attributes could follow the same pattern as the monocle_trace context manager which already accepts an attributes parameter.

Related: Issue #18 (runtime method tracing configuration).

kunalrawat425:

Closes #634


kunalrawat425:

PR #744, opened 2026-07-22
fix: address issue #632 - [Feature Request]: Support reusing trace context for arbitrary Python runtimes

### LINKED ISSUE #632: [Feature Request]: Support reusing trace context for arbitrary Python runtimes
Current behavior: There is no mechanism to propagate or reuse an existing trace context in arbitrary Python runtimes outside of the auto-instrumented frameworks. For example, when initializing Ray cluster workers, each worker starts a new trace context rather than continuing the parent trace.

Expected behavior: Provide an API to pass and reuse an existing trace ID/context in arbitrary Python runtimes, so that spans created in worker processes (e.g., Ray workers) are correlated under the same trace as the parent process.

Where to start: This is a trace context propagation feature. The existing monocle_trace context manager and monocle_trace_method decorator create new traces — a variation is needed that accepts an existing trace context and continues it. Related to issue #18 (runtime method tracing configuration).

kunalrawat425:

Closes #632


kunalrawat425:

Issue #75, opened 2024-11-15
Inconsistencies in cloud exporters


prasad-okahu:
- Base exporter has dependency on Azure package. The Azure specific error handling should be done in the Azure exporter
- S3 exporter has default region us-east-1 and a default container name. We shouldn't hardcode those as default. Those values need to be configured by users.
-

sanskar-singh-2403:
@prasad-okahu can you assign this to me? I have opened a pr to fix this, the base exporter and region issues were already fixed i added the fix for s3 bucket name as well, Thanks!

sanskar-singh-2403:
hey @prasad-okahu requesting your review here so that i can fix the pr if it is not needed, Thanks!

PR #759, opened 2026-07-25
fix(exporters): require explicit bucket/container name in cloud exporters

sanskar-singh-2403:
## Summary

Cloud exporters silently fell back to placeholder names when no bucket or
container was configured (`default-bucket` for S3, `default-container` for
Azure Blob). These are not real resources - with S3 the fallback would try to
auto-create a bucket literally named `default-bucket`, and with Azure it would
target a container the user never intended.

This PR drops the hardcoded defaults and raises a clear `ValueError` when
neither the constructor argument nor the environment variable is set, matching
the existing `GCSSpanExporter` behaviour. This makes the four cloud exporters
consistent, which is the subject of #75.

## Changes

- `S3SpanExporter`, `OpenDALS3Exporter`: require `bucket_name` or `MONOCLE_S3_BUCKET_NAME`
- `AzureBlobSpanExporter`, `OpenDALAzureExporter`: require `container_name` or `MONOCLE_BLOB_CONTAINER_NAME`
- Fix a latent bug in `OpenDALAzureExporter` where `self.container_name` was set from the raw arg before the env-var fallback ran, so the resolved name was never stored

## Notes

The other two points in #75 (base exporter depending on the Azure package, and
the hardcoded `us-east-1` S3 region) appear already resolved on `main` by later
refactors.

## Testing

Instantiated each exporter with and without configuration: missing
bucket/container now raises `ValueError`; with the env var set, construction
proceeds past the name check as expected.

### DIFF
--- a/apptrace/src/monocle_apptrace/exporters/aws/s3_exporter.py
+++ b/apptrace/src/monocle_apptrace/exporters/aws/s3_exporter.py
@@ -46,7 +46,12 @@ class S3SpanExporter(SpanExporterBase):
                 aws_secret_access_key=os.getenv('AWS_SECRET_ACCESS_KEY'),
                 region_name=region_name,
             )
-        self.bucket_name = bucket_name or os.getenv('MONOCLE_S3_BUCKET_NAME','default-bucket')
+        self.bucket_name = bucket_name or os.getenv('MONOCLE_S3_BUCKET_NAME')
+        if not self.bucket_name:
+            raise ValueError(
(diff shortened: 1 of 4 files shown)


sanskar-singh-2403:
@prasad-okahu @pvatokahu can we review this as well? Thanks!

prasad-okahu (approved):

prasad-okahu:
Merged the patch to main. Thanks @sanskar-singh-2403 for your contribution.

PR #760, opened 2026-07-27
feat(test-tools): record test outcome as Okahu label on cleanup

prasad-okahu:
## What

Records a test's pass/fail verdict back to Okahu as an evaluation label when the test's spans come from the Okahu trace source, so the Monocle test framework's verdict is visible alongside the trace in Okahu.

## How

- New `TraceSource` abstraction under `test_tools/src/monocle_test_tools/trace_sources/`:
  - `TraceSource` ABC with an overridable, default no-op `record_test_result(...)`. Kept general so the existing span loaders (`OkahuSpanLoader`, `JSONSpanLoader`) can migrate under it later.
  - `OkahuTraceSource` — owns all Okahu specifics and posts to `POST /v1/eval/label`.
  - `get_trace_source(name)` factory.
- `MonocleValidator`:
  - Captures the loaded fact's id / Okahu fact name / workflow name in `import_traces` (mapping `trace`→`traces`, `session`→`agent_sessions`, `scope`→scope name), reset in `cleanup()`.
  - `post_test_cleanup` calls `_maybe_record_test_result(...)` after flush, before cleanup.

## Behavior

Recorded body: `{"result": {"label": <test_name>, "value": "PASS"|"FAIL", "description": "Test run", "type": "test"}}` with `trace_id` / `fact_name` / `workflow_name` query params.

Recording fires **only** when all hold:
1. trace source is `okahu`,
2. an `OkahuSpanExporter` is active,
3. `OKAHU_API_KEY` is set,
4. a fact id and workflow name are available.

It is **best-effort**: any failure (missing key, network, non-2xx) logs a warning and never fails the test or masks its real result. Works for all facts (trace / session / scope).

## Tests

- 22 new unit tests (`test_okahu_trace_source.py`, `test_validator_record_result.py`): request shape, PASS/FAIL mapping, all gating no-op paths, best-effort failure handling, fact-detail capture, and cleanup routing.
- Full `test_tools` unit suite shows no new failures versus baseline.

### DIFF
--- a/test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
+++ b/test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
@@ -29,7 +29,7 @@ class OkahuTraceSource(TraceSource):
     def record_test_result(self, *, fact_id: Optional[str], fact_name: Optional[str],
                            workflow_name: Optional[str], test_name: str,
                            test_failed: bool, exporters: Optional[list] = None,
-                           description: str = "Test run") -> bool:
+                           description: str = None) -> bool:
         """Record the test outcome as an Okahu eval label.
 
         Best-effort: returns ``False`` (and logs at most a warning) when the
@@ -55,7 +55,7 @@ class OkahuTraceSource(TraceSource):
             "result": {
                 "label": test_name,
                 "value": value,
-                "explanation": description,
+                "explanation": description if description else value,
                 "category": "test",
             }
         }
--- a/test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
+++ b/test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
@@ -55,8 +55,8 @@ class OkahuTraceSource(TraceSource):
             "result": {
                 "label": test_name,
                 "value": value,
-                "description": description,
-                "type": "test",
+                "explanation": description,
+                "category": "test",
             }
         }
 
--- a/test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
+++ b/test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
@@ -1,0 +1,88 @@
+"""Okahu trace source.
+
+Records a test's pass/fail outcome back to Okahu as an evaluation label via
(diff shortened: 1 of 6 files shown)


prasad-okahu:
@karmehr-okahu, please take a look and provide your feedback. Thanks

karmehr-okahu (approved):

karmehr-okahu:
> @karmehr-okahu, please take a look and provide your feedback. Thanks

These changes looked correct and appropriate.

test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if not fact_id or not workflow_name:
+            logger.debug("Missing fact_id or workflow_name; skipping test-result label.")
+            return False
+
+        base = os.getenv("OKAHU_EVALUATION_ENDPOINT", OKAHU_PROD_EVALUATION_ENDPOINT).rstrip("/")
+        value = "FAIL" if test_failed else "PASS"
+        payload = {
+            "result": {
+                "label": test_name,
```
karmehr-okahu, line 56: Why is test_name being stored as the label instead of eval_result?

prasad-okahu: The label ie template name is the test name, pass/fail is the value.

test_tools/src/monocle_test_tools/trace_sources/okahu_trace_source.py
```diff
@@ -0,0 +1,88 @@
...
+
+class OkahuTraceSource(TraceSource):
+    """Trace source backed by the Okahu cloud service."""
+
+    name = "okahu"
+
+    def record_test_result(self, *, fact_id: Optional[str], fact_name: Optional[str],
+                           workflow_name: Optional[str], test_name: str,
+                           test_failed: bool, exporters: Optional[list] = None,
+                           description: str = "Test run") -> bool:
```
karmehr-okahu: For consistency reasons, consider changing the description parameter to explanation to match the other changes. Functionally this won't change anything.

karmehr-okahu (approved):

PR #763, opened 2026-07-27
fix(test-tools): eval reliability — stale trace source, transient retries, and why a row failed

cjames-okahu:
Tracking issues: okahu/monocle_backlog_tracking#222 and okahu/monocle_backlog_tracking#223

Three commits hardening the eval path. All three came out of one 3-run × 20-scenario
hallucination replay matrix in which **10 of 60 evals** did not produce a usable result: one
scenario was lost in every run to a stale trace source, three more to transport failures, and
none of them said why in the recorded matrix.

## 1. `fix`: resolve the trace source independent of fluent call order

The **first** `check_eval` of every pytest session failed, and every later test passed:

```
Unexpected response format from evaluation service. Expected 'result' key in response.
Received: {'message': 'Job submitted', 'job_id': 'interactive_..._1785173415', 'result': []}
```

The message is misleading: `result` **is** present and empty, and the real error is an
`IndexError` on `result[0]` that a blanket `except Exception` relabels.

`MonocleValidator` is a process-wide singleton whose `_trace_source` starts `""`.
`with_evaluation()` snapshots that value into the evaluator at construction, and the CSV adapter
(or any test shaped like `case.run(asserter.with_evaluation("okahu"), ...)`) configures the
evaluation *before* `with_trace_source()` runs. So the session’s first test built an evaluator
with an empty source, and two derived fields were wrong:

| derived | wrong value | consequence |
|---|---|---|
| `shadow_eval = _trace_source != "okahu"` | `True` | the service runs a **shadow** eval, which returns `result: []` by design |
| `_trace_exported = _trace_source == "okahu"` | `False` | re-exports a trace the okahu source already holds → a spurious `401` |

Later tests worked only because the singleton retained `"okahu"` from the first test. Retrying
could not help — a retry faithfully resubmits `shadow_eval=True` (confirmed: 3 retries, all
empty, each minting a new `job_id`). The failure follows position, not trace:
`test_cc_t02`/`test_cc_t03` each fail this way when run first and pass when run second.

Fix: a `BaseEval.set_trace_source()` hook, an `OkahuEval` override keeping `_trace_source` and
`_trace_exported` in step, and `with_trace_source()` pushing the real source into an
already-constructed evaluator. Call order no longer changes how an eval is submitted.

## 2. `feat`: record why an eval produced no label

A failed row carried `actual: ""` and an empty `explanation`, so the cause lived only in the
pytest log — which is precisely why the bug above read as "a bad trace" for hours.
`check_eval` now stashes the exception message at the raise site and the matrix row exposes an
additive `failure_reason` column. Empty for passes and for judge disagreements (already
described by `actual` + `explanation`); existing consumers unaffected.

## 3. `fix`: retry transient failures without duplicating eval jobs

Three further scenarios were lost and none was retried: `get_fact_map()` had no retry at all,
and the submit POST retried only `requests.Timeout`, so `ConnectionResetError` (a
`RequestException`) fell straight through.

Retrying is only safe where re-issuing cannot change state. Submitting is **not** idempotent —
every POST creates a new job id and a new charge, confirmed by observation (identical params
returned four different job ids). So two policies:

- **idempotent reads** (fact map, template listing) — retry any transport failure with
  exponential backoff.
- **the submit POST** — retry **only** `requests.ConnectTimeout`, where the connection was never
  established so nothing reached the service.

This deliberately **narrows** the submit path, which previously retried `ReadTimeout` — requests
that *were* sent — a pre-existing double-charge exposure. Those now fail fast with a message
saying why no retry was attempted, so an operator can distinguish "never ran" from "may have
run" (and with commit 2 that reason lands in the matrix). Say the word if you would rather
recover those at the risk of duplicate jobs and I will widen `_SUBMIT_RETRY_ON`.

`_submit_eval_job()` is untouched — it has no callers.

## Tests

18 new tests, each driven red→green:

- `test_eval_trace_source_ordering.py` (4) — `shadow_eval=False` and no re-export for the
  eval-configured-first order, plus two controls: the working order keeps working, and a `file`
  source still correctly shadow-evals (guards against over-correcting).
- `test_eval_matrix_failure_reason.py` (5) — schema field, empty default, capture at the raise
  site, judge mismatches still described by `explanation`, JSON round-trip.
- `test_okahu_eval_transient_retry.py` (9) — both policies, growing backoff, no sleep on
  success, and two **safety** tests: a reset and a read timeout on submit must each produce
  exactly ONE POST.
- Updated `test_build_eval_matrix_row_pass_includes_tokens`, an exact-schema snapshot, for the
  additive key.

Full `test_tools/tests/unit`: **350 passed** (`test_non_llm_core_examples.py` excluded —
pre-existing `ModuleNotFoundError: google.adk`, unrelated).

## Live verification

Against the prod eval service, the first scenario of a session:

| | before | after |
|---|---|---|
| `shadow_eval` sent | `True` | `False` |
| `result` length | 0 | 1 |
| outcome | fail | pass |
| trace-exporter 401 | yes | no |
| 2-scenario runtime | 101s | 34s |

Retry verified separately against a refused endpoint: the fact-map read now makes 3 attempts
with 1s/2s backoff instead of failing instantly. And the **combined** branch re-verified live:
first-of-session scenario `shadow=False`, judged, 5,901 tokens, `failure_reason` empty on a pass,
no 401.

## Note

On a hard outage `get_fact_map()` retries once per call, because its cache only populates on
success — so a scenario can spend two backoff cycles. Harmless; left alone to keep the change
narrow.

---

Supersedes #764, which contained commit 3 alone and is now closed in favour of this PR.

### DIFF
--- a/test_tools/src/monocle_test_tools/eval_matrix.py
+++ b/test_tools/src/monocle_test_tools/eval_matrix.py
@@ -29,15 +29,17 @@ def build_eval_matrix_row(run_id: str, scenario: str, last_eval: dict, passed: b
     `last_eval` is expected to look like the dict stashed by
     `TraceAssertion.check_eval`:
         {"trace_id": str, "expected": ..., "fact_name": str, "label": ...,
-         "explanation": ..., "judge_output": dict, "total_tokens": ...}
+         "explanation": ..., "judge_output": dict, "total_tokens": ...,
+         "failure_reason": str}
 
     Returns a dict with exactly these keys (schema is consumed downstream,
     do not rename): run_id, scenario, trace_id, expected, actual, status,
     explanation, total_tokens, claim_verdicts, hallucination_types,
-    entity_match_check, fact_id, workflow, job_id.
+    entity_match_check, fact_id, workflow, job_id, failure_reason.
 
     The trailing fact_id/workflow/job_id are additive columns for the filtered
     flow; interactive rows carry empty-string defaults so no consumer breaks.
+    `failure_reason` is likewise additive and empty unless evaluate() raised.
     """
     last_eval = last_eval or {}
     judge_output = last_eval.get("judge_output") or {}
@@ -65,6 +67,10 @@ def build_eval_matrix_row(run_id: str, scenario: str, last_eval: dict, passed: b
         "fact_id": last_eval.get("fact_id") or "",
         "workflow": last_eval.get("workflow") or "",
         "job_id": last_eval.get("job_id") or "",
+        # Why the eval produced no label, when it produced none. Empty for passing
+        # rows and for judge disagreements (those are described by actual +
+        # explanation); populated when evaluate() raised.
+        "failure_reason": last_eval.get("failure_reason") or "",
     }
 
 
--- a/test_tools/src/monocle_test_tools/evals/base_eval.py
+++ b/test_tools/src/monocle_test_tools/evals/base_eval.py
@@ -14,6 +14,16 @@ class BaseEval(BaseModel):
         """Optional cleanup hook called at test end. Override if needed."""
         pass
 
+    def set_trace_source(self, trace_source: str) -> None:
+        """Adopt a trace source discovered after this evaluator was constructed.
+
+        `with_evaluation()` may run before `with_trace_source()` -- e.g.
+        `case.run(asserter.with_evaluation("okahu"), ...)`, where the evaluation is
+        configured while building the argument -- so the source is not always known
+        at construction time. Override if the evaluator's behaviour depends on it.
+        """
+        pass
+
     @classmethod
     def classify_eval_input(cls, name_or_path: str) -> Tuple[str, str]:
         """Classify an eval input as builtin or custom.
--- a/test_tools/src/monocle_test_tools/evals/okahu_eval.py
+++ b/test_tools/src/monocle_test_tools/evals/okahu_eval.py
@@ -54,7 +54,19 @@ class OkahuEval(BaseEval):
         self._fact_map_cache = None
         self.last_judge_output: dict = {}
         self.last_total_tokens = None
-    
+
+    def set_trace_source(self, trace_source: str) -> None:
+        """Adopt a trace source configured after construction.
+
+        Both `shadow_eval` and `_trace_exported` derive from `_trace_source`, so a
+        stale value here means the eval is submitted as a shadow eval (the service
+        then returns an empty `result`) and the trace is needlessly re-exported.
+        Keep the two in step whenever the source changes.
+        """
+        self._trace_source = trace_source or ""
+        # An okahu source is already holding the trace, so there is nothing to export.
+        self._trace_exported = self._trace_source == "okahu"
+
     @staticmethod
     def _map_fact_name(fact_name: str) -> str:
         """
--- a/test_tools/src/monocle_test_tools/evals/okahu_eval.py
+++ b/test_tools/src/monocle_test_tools/evals/okahu_eval.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import time
 from datetime import datetime, timezone
 from opentelemetry.sdk.trace import Span
 import requests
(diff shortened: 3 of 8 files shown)


prasad-okahu:
@shambhavigoel-13 , please take a look and see how much of the changes you recently added covers this.

cjames-okahu:
Pushed a follow-up commit (`ef6eec4`) that fixes a **duplicate-write regression this PR would otherwise introduce** — flagging for review since it's a direct consequence of the trace-source ordering fix.

### What happens
`OkahuEval.evaluate()` client-exports the eval result whenever `"okahu" in MONOCLE_EXPORTER or self._trace_source == "okahu"`. An okahu trace source is evaluated **non-shadow** (`shadow_eval=False`), so the service already persists the result server-side. The `self._trace_source == "okahu"` disjunct then fires a **second** `POST /v1/eval/jobs/{job_id}/results`, writing a duplicate `eval_results` row for the same `(job_id, id)`.

That disjunct predates this branch but was **dormant**: before the ordering fix the evaluator usually saw an empty `_trace_source`, so the gate was `False` and the client export never ran. Once ordering reliably reports `"okahu"`, the gate is `True` for every okahu-source eval — so this PR, correctly, turns the dormant path on.

### Evidence (shared prod tenant, same workflow)
- **interactive** eval `job_id`s: **~6% → ~96%** of `(job_id, trace)` pairs doubled across the ordering change; two byte-identical rows, same `job_id`, ~2–3s apart.
- **batch** (`okahueval_`) `job_id`s, which never client-export: **0 / 8,304** doubled.

### Fix
Gate the client export solely on `MONOCLE_EXPORTER` containing `"okahu"` — the explicit opt-in the comment already describes. This mirrors the trace-export short-circuit (`_trace_exported` is `True` for an okahu source because the trace already lives in okahu); the result already lives there too. The `MONOCLE_EXPORTER=okahu` opt-in path is unchanged.

Regression tests added to `test_eval_trace_source_ordering.py`: an okahu-source eval must not client-export its result (both fluent call orders); `MONOCLE_EXPORTER=okahu` still does. Verified they fail on the pre-fix code and pass after. Full unit suite: 300 passed (the one collection error, `test_non_llm_core_examples.py`, pre-exists on `main` — missing optional `google.adk`).

One thing I could **not** confirm from the client side: that the sync submit's server-side persist is unconditional. The batch path proves the server persists without any client export, and the two-writer timing fits, but if there's a config where the sync `POST /v1/eval/jobs` does *not* persist, this fix would need pairing with a server-side idempotent upsert on `(job_id, id)`. Worth a sanity check against the eval-api service.

cjames-okahu:
### Verified end-to-end against a live tenant

Built a wheel from this branch and ran 4 real interactive hallucination evals (okahu trace source, `MONOCLE_EXPORTER` unset) against prod. Then queried `eval_results` for those 4 traces in the run window:

| trace | rows written | label |
|---|---|---|
| `f68634…d507f` | **1** | no_hallucination |
| `e653ee…fb02`  | **1** | major_hallucination |
| `3fefdc…105ed` | **1** | no_hallucination |
| `4c2678…af8d`  | **1** | major_hallucination |

**4/4 wrote exactly one row.** The same traces doubled at ~96% on the unfixed wheel — so the duplicate write is gone.

This also settles the open question from my previous comment (whether the sync submit's server-side persist is unconditional): **it is.** With the client-side `export_results` removed, the service still persisted every eval, with correct labels — so the fix drops the redundant write without dropping data. No server-side idempotent-upsert change is needed for this path.

cjames-okahu:
Rebased onto latest `main` and force-pushed (was conflicting).

While rebasing I **dropped this branch's `stop double-writing eval_results` commit**: that same fix landed on `main` separately via #771, in a stricter form (it gates the client-side result export on `shadow_eval` as well as `MONOCLE_EXPORTER`). Keeping our version would have been a redundant, weaker duplicate, so the branch now defers to `main` for that behavior.

What remains here is the original three fixes only — trace-source ordering, the additive `failure_reason` matrix column, and the transient-retry policy. The ordering fix composes correctly with #771's guard: an okahu source is evaluated non-shadow, so `main`'s `shadow_eval and ...` gate suppresses the client export and no duplicate row is written.

Full `test_tools/tests/unit`: 350 passed.

Note: #771 shipped without a regression test for the duplicate-write behavior, and dropping our commit removed the export-count tests that had covered it. Happy to add a standalone regression test to `main` for that guard in a separate PR if useful.

shambhavigoel-13:
Hi @prasad-okahu  Seems like a new change of code that is needed has only been implemented here. Doesn't seem any recent changes are associated with these.

PR #764, opened 2026-07-27
fix(test-tools): retry transient eval-service failures without duplicating eval jobs

cjames-okahu:
Tracking issue: okahu/monocle_backlog_tracking#223

## Problem

A dropped connection to the eval service costs a whole scenario. On a 3-run × 20-scenario
replay matrix, 3 of 60 evals were lost and none was retried:

| scenario | failure | why no retry |
|---|---|---|
| `cc_t13` | `ConnectionResetError(54)` on submit | a reset is a `RequestException`; the submit loop caught only `requests.Timeout` |
| `cc_t15` | `ConnectionResetError(54)` on submit | same |
| `cc_t17` | `Fact map request timed out (read timeout=60)` | `get_fact_map()` had no retry at all |

## The constraint

Submitting an eval job is **not idempotent**: every POST to `/v1/eval/jobs` creates a new
job id and a new charge. Confirmed by observation — replaying identical params returned
four different job ids (`..._1785176597`, `..._608`, `..._617`, `..._646`). A blind retry
would double-bill and corrupt token accounting.

## Two policies

- **`_IDEMPOTENT_RETRY_ON`** — fact map and template listing. Retry any transport failure
  (`ConnectionError`, `Timeout`) with exponential backoff. Re-reading is free.
- **`_SUBMIT_RETRY_ON`** — the eval-job POST. Retry **only** `requests.ConnectTimeout`,
  where the connection was never established, so nothing reached the service.

### This narrows existing behaviour, on purpose

The submit path previously retried `requests.Timeout`, which includes `ReadTimeout` —
requests that *were* sent. That was a pre-existing double-charge exposure. Now a read
timeout or reset on submit fails immediately, and the message says why it was not retried:

```
Evaluation service did not respond in time: ... Not retried, because the job was already
submitted and resubmitting would create a duplicate eval job.
```

so an operator can distinguish "never ran" from "may have run". If you would rather keep
recovering those at the risk of duplicate jobs, say so and I will widen `_SUBMIT_RETRY_ON`.

`_submit_eval_job()` is untouched — it has no callers.

## Tests

9 new tests in `test_okahu_eval_transient_retry.py`, driven red→green:

- fact map retries a reset, and a read timeout, then succeeds
- fact map stops at the attempt budget and still raises its usual message
- backoff grows between attempts; a successful read never sleeps
- submit retries a `ConnectTimeout` then succeeds, and stops at the budget
- **safety:** a reset on submit produces exactly ONE POST
- **safety:** a read timeout on submit produces exactly ONE POST

Full `test_tools/tests/unit`: **288 passed** (`test_non_llm_core_examples.py` excluded —
pre-existing `ModuleNotFoundError: google.adk`, unrelated).

## Live verification

Against a refused endpoint, the fact-map read now makes 3 attempts with 1s/2s backoff
(11.5s) instead of failing instantly (2.4s).

## Note

On a hard outage `get_fact_map()` retries once per call, because the cache is only
populated on success — so a scenario can spend two backoff cycles (~6s) rather than one.
Harmless, and left alone to keep this change narrow.

## Relationship to #763

Independent, and based on `main` rather than that branch. #763 fixes a stale trace source
(and records `failure_reason`); this hardens the transport layer. They compose: with both,
a transient failure is retried where safe, and when it is not recoverable the reason lands
in the eval-result matrix instead of only the pytest log.

### DIFF
--- a/test_tools/src/monocle_test_tools/evals/okahu_eval.py
+++ b/test_tools/src/monocle_test_tools/evals/okahu_eval.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import time
 from datetime import datetime, timezone
 from opentelemetry.sdk.trace import Span
 import requests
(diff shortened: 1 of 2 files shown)


cjames-okahu:
Rolled into #763 at the reviewer's request, so the eval-reliability work lands as one change.

Commit 47f3216 was cherry-picked onto `fix/eval-trace-source-ordering` as 5035b04 (clean auto-merge, no conflicts) and #763 now carries all three commits. Nothing is lost — the retry policies, the two safety tests, and the deliberate `ReadTimeout` narrowing are all in #763, whose description now covers them.

Verified in the combined state: 297 unit tests pass, all 18 new tests pass together, and a live first-of-session run gave shadow_eval=False with a judged label and no 401.

Tracking issue okahu/monocle_backlog_tracking#223 now points at #763.

PR #778, opened 2026-08-10
feat(exporters): add ClickHouse span exporter

hocokahu:
## What this adds

A ClickHouse span exporter, so Monocle traces can be written to a ClickHouse `traces` table. It works like any other exporter:

```bash
pip install "monocle-apptrace[clickhouse]"
export MONOCLE_EXPORTER=clickhouse
export MONOCLE_CLICKHOUSE_CONNECTION_URL="clickhouse://user:pass@host:8123/db"
```

## How it's built

`ClickHouseSpanExporter` subclasses `SpanExporterBase` and reuses the shared helpers (`serialize_span`, the `0x` span/trace id formatting, `skip_export`). It writes one row per span into a `traces` table:

- Client is `clickhouse-connect` over HTTP.
- Table uses `MergeTree ORDER BY (trace_id, span_id)`. ClickHouse has no autoincrement key and needs a sort key.
- JSON fields (`status`, `attributes`, `events`) are stored as `String` columns holding `json.dumps(...)`.
- `parent_id` and `metadata` are `Nullable(String)`; timestamps are `DateTime64(9, 'UTC')`.
- On a dropped connection it reconnects and retries the insert once. A missing-privilege error surfaces as ClickHouse code 497 (`ACCESS_DENIED`) and is raised as a `PermissionError`.

## Not a new required dependency

`clickhouse-connect` is an optional extra under `[project.optional-dependencies]`, like the other optional exporter extras (`aws`, `azure`, `gcp`). Users who don't set `MONOCLE_EXPORTER=clickhouse` install nothing extra and never import it, because the registry imports the exporter module lazily.

## What changed

- New exporter under `apptrace/src/monocle_apptrace/exporters/clickhouse/`
- Registered `clickhouse` in the exporter registry
- New `clickhouse` extra in `apptrace/pyproject.toml`
- Added `clickhouse` to the unit-test CI install, so the exporter's unit tests run in CI instead of skipping on a missing optional dependency

## Tests

Unit tests use a mocked client and need no server:

```bash
pip install "monocle-apptrace[clickhouse]"
pytest apptrace/tests/unit/test_clickhouse_exporter.py
# 18 passed
```

The integration test runs against a real server. The single static ClickHouse binary is enough, no Docker:

```bash
./clickhouse server &        # HTTP on :8123
export MONOCLE_CLICKHOUSE_CONNECTION_URL="clickhouse://default:@localhost:8123/default"
pytest apptrace/tests/integration/test_clickhouse_exporter_integration.py
# 1 passed: "Verified 2 span(s) match between ClickHouse and .monocle for trace 0x..."
```

What the tests assert:

- Unit (18 cases): the exporter registers under the `clickhouse` key; it reads the connection URL from `MONOCLE_CLICKHOUSE_CONNECTION_URL` and raises when it is missing; each built row has `0x`-prefixed ids, timezone-aware timestamps, JSON stored as strings, and null metadata; a batch produces one `insert` call; non-Monocle spans are skipped; a serialization error on one span does not drop the others; a dropped connection triggers one reconnect and retry; shutdown closes the client; init creates the table; a privilege error becomes a `PermissionError`.
- Integration: it exports one trace (a root span and a child span) through both the ClickHouse exporter and the file exporter, reads the rows back from the `traces` table and the spans from the `.monocle` file, and asserts they describe the same trace: the same set of span ids, and for each span the same `name`, `parent_id`, `status`, `attributes`, and `events`.

To look at the rows directly after the integration run:

```bash
./clickhouse client --query "SELECT name, span_id, parent_id FROM traces FORMAT PrettyCompact"
```

### DIFF
--- a/apptrace/tests/integration/test_clickhouse_exporter_integration.py
+++ b/apptrace/tests/integration/test_clickhouse_exporter_integration.py
@@ -1,16 +1,19 @@
-"""End-to-end check that ClickHouseSpanExporter persists the same span data
-that FileSpanExporter writes to the local ``.monocle`` folder.
+"""End-to-end check that ClickHouseSpanExporter writes spans into the new
+schema and that the data is queryable there:
 
-The same spans are exported through both exporters, then the rows read back
-from ClickHouse are compared field-by-field against the JSON written to disk.
+- attributes  -> JSON        (sub-fields addressable, e.g. attributes.`span.type`)
+- events      -> Array(JSON)  (searchable per element, e.g. by event name)
+- status      -> status_code + status_message columns
+
+The stored shape intentionally differs from the local ``.monocle`` file (the
+JSON type re-nests dotted keys and status is split into columns), so this test
+verifies the ClickHouse content directly rather than byte-for-byte parity.
 
 Requires a running ClickHouse server reachable via
-``MONOCLE_CLICKHOUSE_CONNECTION_URL`` (a single-binary ``clickhouse server``
-is enough); the test skips itself otherwise.
+``MONOCLE_CLICKHOUSE_CONNECTION_URL`` (a single-binary ``clickhouse server`` is
+enough); the test skips itself otherwise.
 """
 import os
-import glob
-import json
 import logging
 
 import pytest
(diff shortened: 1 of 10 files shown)


prasad-okahu (commented):
@hocokahu, Question on the schema, can we make the events and attributes better searchable? Is there an alternative datatype (eg json) that would help the sub field search?

hocokahu:
Yes, we can use JSON (attributes, metadata) and Array(JSON) for events. I would put status in two columns status_code and status_message since most of the data is stored that way. Thoughts?

prasad-okahu:
> Yes, we can use JSON (attributes, metadata) and Array(JSON) for events. I would put status in two columns status_code and status_message since most of the data is stored that way. Thoughts?

Sounds good. Perhaps align it with schema of Postgres exporter exactly, just with compatible data types. That will make things portable for migration and queries.

hocokahu:
Should the ClickHouse schema mirror Postgres exactly, or use ClickHouse-native types where they help?

The one real structural difference is `events`. ClickHouse `Array(JSON)` lets us filter and index into `events[]` directly, e.g. `arrayFilter(x -> x.name='data.input', events)[1].attributes.input`. Plain `JSON` can't hold a top-level array at all (it errors on insert), so `Array(JSON)` is the only option that works. Postgres `JSONB` stores the same array without error, but you need `jsonb_array_elements()` to unnest it on read.

Otherwise the PR stays consistent with Postgres:

- Added the `metadata` column. It's stored as empty JSON `{}` because a ClickHouse `JSON` column can't be NULL, which matches Postgres's reserved/NULL `metadata` (also unused by the other exporters).
- Split `status` into `status_code` and `status_message`. Status almost always carries both, and separate columns make status filtering and search cheaper. Happy to keep it as a single `status` object to match Postgres if reviewers prefer. Thoughts?

Separate follow-up, not in this PR: using ClickHouse's native features more. For large or base64 payloads we could store a pointer instead of the full input/output, and keep a full-vs-truncated split for span-list and span-detail views.

apptrace/tests/integration/test_clickhouse_exporter_integration.py
```diff
@@ -0,0 +1,89 @@
...
+    conn_url = os.getenv("MONOCLE_CLICKHOUSE_CONNECTION_URL")
+    if not conn_url:
+        pytest.skip("MONOCLE_CLICKHOUSE_CONNECTION_URL not set")
+
+    provider = TracerProvider()
+    provider.add_span_processor(SimpleSpanProcessor(ClickHouseSpanExporter()))
+    tracer = provider.get_tracer(__name__)
+
+    with tracer.start_as_current_span("clickhouse.workflow") as root:
+        root.set_attribute(MONOCLE_SDK_VERSION, "0.8.0")
```
prasad-okahu: Nit: Can we use get_monocle_version() instead of hardcoding 0.8.0

prasad-okahu (commented):
Looks fine to me. Just a small feedback inline.

apptrace/tests/integration/test_clickhouse_exporter_integration.py
hocokahu: Done. Both use `get_monocle_version()` now.

prasad-okahu (approved):

prasad-okahu:
Merged the PR. Thanks @hocokahu !

PR #784, opened 2026-08-14
feat: allow custom spans to exclude inputs and/or outputs (closes #782)

### LINKED ISSUE #782: Add option to Monocle custom span to include inputs and outputs
Monocle has an ability to add a simple span for your custom method by either explicitly annotating the code or by adding that method signature in a configuration file.
Currently, that custom span includes arguments passed to the method and the value returned by that method. We want make the these inputs and outputs optional by setting an optional exclude=inputs|outputs

dchaudhari7177:
Closes #782.

Custom spans always captured the method's arguments and return value, with no way to opt out — which rules out tracing any method that handles credentials, PII, or payment details, since the only alternative was not tracing it at all.

## The option

`exclude` accepts `"inputs"`, `"outputs"`, or both, on **both** surfaces that create custom spans:

```python
@monocle_trace_method(exclude="inputs")
def charge_card(card_number: str) -> Receipt: ...
```

```yaml
instrument:
  - package: myapp.billing
    class: Billing
    method: charge
    exclude: [inputs, outputs]
```

`build_custom_span_processor(exclude)` assembles the processor with the excluded accessors **left out entirely**, so excluded data is never serialized in the first place rather than captured and then dropped.

## Two judgement calls worth reviewing

**Excluding outputs still records `error_code`.** That attribute isn't function output — it's whether the call failed. Dropping it would leave an excluded span unable to say anything went wrong, which seemed clearly against the point of tracing. Excluding *both* therefore leaves a span carrying only `error_code`, which still tells you the method ran and whether it succeeded.

**An unrecognised name warns and is ignored** rather than raising. A typo in a tracing option shouldn't take down the application being traced. Happy to make it strict if you'd rather.

`CUSTOM_SPAN_PROCESSOR` is still exported and still captures both, so nothing importing it changes behaviour — there's a test asserting exactly that.

Normalization accepts a bare string or any iterable, and is case- and whitespace-insensitive, so a value from YAML behaves like one written in code.

## Tests

17 in `tests/unit/test_custom_span_exclude.py`:

- normalization: `None`/empty, bare string, iterable, mixed case and padding, unknown name warns and is dropped
- processor shape for each of the four combinations, and the module default unchanged
- **the decorator end to end** against an `InMemorySpanExporter`: with `exclude="inputs"` the secret argument is absent from the span; with `exclude="outputs"` the return value is absent
- an excluded span is **still emitted**, and still reports `error_code` when the method raises
- the wrapped function's return value is unchanged
- **the YAML config path**, asserting the secret string appears nowhere in the serialized span events

## Suite status

Windows / Python 3.12. `tests/unit` goes **546 → 563 passed** (my 17).

The 4 failures and 15 collection errors are **identical before and after** this branch — I ran the suite on pristine `main` to confirm. The collection errors are missing optional framework deps in this environment (`torch`, `flask`, `langchain_core`, `boto3`, `haystack`, `pydantic`, `starlette`, `httpx`).

One note in case it's useful: `pip install -e apptrace` fails on Windows because the hatchling build hook shells out to `if [ -f "LICENSE" ]; ...`, which `cmd.exe` can't parse. I ran the suite with `PYTHONPATH=src` and the runtime deps installed directly instead.

Signed off per the DCO.

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/common/custom_span_processor.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/common/custom_span_processor.py
@@ -77,11 +77,62 @@ def extract_output(arguments):
         return json.dumps({"error": "Failed to capture output"})
 
 
-# Custom span entity processor following Monocle's pattern
-CUSTOM_SPAN_PROCESSOR = {
-    "type": "custom",
-    "events": [
-        {
+EXCLUDE_INPUTS = "inputs"
+EXCLUDE_OUTPUTS = "outputs"
+VALID_EXCLUSIONS = (EXCLUDE_INPUTS, EXCLUDE_OUTPUTS)
+
+
+def normalize_exclusions(exclude=None):
+    """Normalize an ``exclude`` option into a set of known exclusion names.
+
+    Accepts None, a single string ("inputs"), or any iterable of strings
+    (["inputs", "outputs"]). Comparison is case-insensitive and whitespace is
+    stripped, so values coming from a YAML config behave the same as values
+    passed in code.
+
+    An unrecognised name is warned about and ignored rather than raising:
+    a typo in a tracing option should not take down the application being
+    traced.
+    """
+    if not exclude:
+        return frozenset()
+
+    items = [exclude] if isinstance(exclude, str) else list(exclude)
+    normalized = set()
+
+    for item in items:
+        name = str(item).strip().lower()
+        if name in VALID_EXCLUSIONS:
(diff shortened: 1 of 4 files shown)


prasad-okahu (approved):
Looks fine to me. Thanks for your contribution.
It would be great if you can add a code snippet and configuration change as examples in the README. Feel free to create a follow up PR for that.

prasad-okahu:
@dchaudhari7177 merged the PR. Thanks for your contribution!

PR #791, opened 2026-08-18
feat(test_tools): check_eval takes a built-in eval name or a custom t…

shambhavigoel-13:
…emplate path

Okahu evals come in two flavours -- a built-in template, named by a plain string ("hallucination"), and a custom template, identified by the path of a JSON file holding it. check_eval exposed those as two separate selectors, so callers had to switch parameters (eval_name vs template_path) depending on the flavour, which made it awkward to drive both kinds from one common place (a CSV column, a generated test, a parametrized fixture).

eval_name now accepts either. Which one it holds is decided by its datatype -- a path object (pathlib.Path) is a custom template file, a str is a built-in template name -- and the new keyword-only eval_type ("builtin" / "custom") states it explicitly, which is also how a custom template path is passed as a plain str.

Only the datatype is examined; the string is never sniffed for a ".json" suffix or a separator, so a built-in template whose name happens to look path-like still resolves as a built-in rather than being silently read off disk.

A resolved custom template is routed into the existing template_path slot before check_eval splits into span mode and time-window (filtered) mode, so both flows reuse the current custom-template handling unchanged: file load, {"template": {...}} unwrap, and the name-field fallback for the eval name. template_path and the inline template dict keep working exactly as before.

Contradictory input raises ValueError: an unknown eval_type, eval_type without eval_name, eval_type="custom" alongside template_path/template, and a path object tagged eval_type="builtin" (which would otherwise ship a filesystem path to the eval service as a template name).

## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments


prasad-okahu (commented):
@shambhavigoel-13 , let's not add eval_type as a new parameter. That defeats the purpose of making is seamless.
The `_apply_eval_type()` should detect the type for eval_name, if this Path object or a string like path, then the type is Custom eval. The returns from _apply_eval_type() should return the right combination of name and path the for downstream check_eval to work correctly.

prasad-okahu (approved):

PR #792, opened 2026-08-18
fix(instrumentation): skip health check spans that return a response body

shambhavigoel-13:
## Problem

`HttpSpanHandler.should_sample` only recognised a `GET`/`HEAD` span as a health check when its output was **completely empty**. Real health endpoints answer with a payload (`{"status":"ok"}`, `OK`), so `response is not None` classified every one of them as real traffic and exported it. With health checks running every few seconds this fills the trace with health check spans.

A second, separate bug: the failure check read only `error_code`, which is emitted by the lambda and agentcore metamodels alone. FastAPI, Flask, aiohttp and azfunc report `status_code`, so for those the branch was dead code and a **failing** health check with an empty body was sampled away 99 times out of 100.

## Fix

- `is_health_check_route()` — a span whose route (or url path) is a well known health check path may carry a response body and still be sampled down to one span per `HEALTH_RESET_COUNTER`. Matches exactly or as a suffix, so `/actuator/health` and `/api/v1/healthz` are covered. Defaults live in `HTTP_HEALTH_CHECK_ROUTES` and can be replaced with `MONOCLE_HEALTH_CHECK_ROUTES`.
- `has_error()` — reads both `error_code` and `status_code`, plus the span status and `MONOCLE_DETECTED_SPAN_ERROR`, and treats any status below 400 as successful. Failing health checks are always exported.

Request params or a request body still force an export, so a real caller hitting `/healthz?verbose=1` is never dropped.

The full gate a span must pass to be skipped: no exception and sampling enabled -> `GET`/`HEAD` -> no `data.input` attributes -> no error -> empty response **or** health check route -> 1 in `HEALTH_RESET_COUNTER` counter.

## Verification

Measured against a live FastAPI app, 3 requests to each endpoint:

| endpoint | before | after |
|---|---|---|
| `/healthz` -> `{"status":"ok"}` | 3 exported | 1 |
| `/livez` -> `OK` | 3 | 1 |
| `/readyz` -> 503, empty body | 1 | **3** |
| `/hello` -> `{"Status":"Success"}` | 3 | 3 |
| `/health/detail` (not a health check route) | 3 | 3 |
| `/healthz?verbose=1` | 3 | 3 |

New unit tests in `apptrace/tests/unit/test_http_span_handler_health_check.py`: 22 pass on this branch, 13 of them fail on `main`.

Note: `test_verify_skip_health_check` in the Flask and aiohttp integration suites fails when those suites run together. Verified it fails identically on unmodified `main` — a pre-existing ordering flake, not from this change.

## Docs

Added a "Health check sampling" section to `docs/Monocle_User_Guide.md` covering `MONOCLE_SAMPLE_HEALTH_CHECKS` and the new `MONOCLE_HEALTH_CHECK_ROUTES`.


prasad-okahu (approved):
@shambhavigoel-13 , fix look fine. The commit is missing DCO. Please add that and I'll merge it. Thanks

shambhavigoel-13:
Please take a look @prasad-okahu Done the dco

prasad-okahu (approved):

PR #798, opened 2026-08-21
fix(instrumentation): populate tool.name/tool.type for ReAct-style LiteLLM tool calls

### LINKED ISSUE #797: LiteLLM ReAct-style tool calls (CrewAI) leave tool.name/tool.type unpopulated
## What's happening

For LiteLLM-backed agents that use ReAct-style text tool calling (CrewAI's default prompting, for example), Monocle correctly reclassifies the inference span as a tool call, but the span's `tool.name` and `tool.type` attributes stay `None`.

PR #371 taught `extract_finish_reason()` (`litellm/_helper.py`) to recognize a `"\nAction: <tool>"` line in `message.content` and return `finish_reason="tool_calls"` for it, since CrewAI signals a tool invocation as text instead of a native `tool_calls` entry. `extract_tool_name()` and `extract_tool_type()` were not updated the same way: they only read `response.choices[0].message.tool_calls`, which is absent for this response shape, so both return `None` even though the span is already typed as a tool call.

## Reproduction

Run against current `main` (`0e1dcc7`):

```python
from types import SimpleNamespace
from monocle_apptrace.instrumentation.metamodel.litellm._helper import (
    extract_finish_reason, extract_tool_name, extract_tool_type,
)

message = SimpleNamespace(
    content="Thought: I should look this up\nAction: search_tool\nAction Input: {}",
    tool_calls=None,
)
choice = SimpleNamespace(finish_reason="stop", message=message)
response = SimpleNamespace(choices=[choice])
arguments = {"result": response, "exception": None}

print(extract_finish_reason(arguments))  # "tool_calls"
print(extract_tool_name(arguments))      # None -- expected "search_tool"
print(extract_tool_type(arguments))      # None
```

## Expected

`extract_tool_name`/`extract_tool_type` should agree with `extract_finish_reason` and report the tool parsed from the `Action:` line when there is no native `tool_calls` entry.

I have a fix and a regression test ready and will open a PR linked to this issue.

AmirF194:
## Root cause

`extract_finish_reason()` (`litellm/_helper.py`) already recognizes a `"\nAction: <tool>"` line in `message.content` and reclassifies the response as `finish_reason=tool_calls`, since CrewAI's ReAct-style prompting signals a tool call as text instead of a native `tool_calls` entry. `extract_tool_name()`/`extract_tool_type()` were never given the same handling: they only look at `response.choices[0].message.tool_calls`, which does not exist for this response shape, so both keep returning `None` even though the span's `subtype` is already `INFERENCE_TOOL_CALL`.

## Fix

Factored the `Action:` line parsing (including the existing `final answer`/`i now know` termination filter) out of `extract_finish_reason` into `_extract_react_action_name`, and reuse it in `extract_tool_name` as the fallback when `_get_first_tool_call` finds no native tool call. `extract_finish_reason`'s own behavior is unchanged, just delegated to the shared helper.

## Verified

- Reproduced against current `main` (`0e1dcc7`) in a clean container: `extract_finish_reason` returns `tool_calls` while `extract_tool_name`/`extract_tool_type` return `None` for a ReAct-style response.
- New regression tests in `apptrace/tests/unit/test_litellm_helper.py` (`TestLiteLLMReactStyleToolName`) fail on `main` and pass on this branch: ReAct tool call resolves a name, the existing native `tool_calls` path still resolves correctly, the `final answer`/plain-text termination cases still return `None`, and the live `INFERENCE` metamodel's `tool.name`/`tool.type` accessors agree.
- `apptrace/tests/unit/test_litellm_helper.py` and `test_litellm_responses_toolcall.py` (the only two unit test files touching this module) pass in full: 38/38.
- Checked every production caller of `extract_tool_name`/`extract_tool_type`/`extract_finish_reason` in `apptrace/src`: the LiteLLM ones are used only by `litellm/entities/inference.py`; every other module with a same-named function is a separate, unrelated metamodel.
- Not verified: no live CrewAI+LiteLLM call was made; the fix and tests exercise the exact response shape `extract_finish_reason` already contracts for.

Fixes #797


prasad-okahu (approved):
@AmirF194  Looks fine to me. Thanks for your contribution.

prasad-okahu:
@AmirF194,  I'll rebase the branch and merge it once the test run is clean.

prasad-okahu:
@AmirF194, Merged PR to main. Again, thanks for your contribution!

AmirF194:
Thanks for the quick rebase and merge. Appreciate you taking it through.

PR #806, opened 2026-08-28
Support parametric tests with fluent API for eval tuning and A/B testing 

prasad-okahu:
## Proposed changes

# Test cases as data: drive the fluent API from a `FluentTestCase`

Writing a Monocle validation test today means restating, call by call, what a test case already describes. A parametrized dict carries the input, the agents, the tools and the expected eval results — and the fluent API takes none of it, so the author unpacks it by hand and the two drift apart.

This adds `testcase=` to the fluent API. One dict drives the run and every assertion:

```python
TESTCASE = {
    "input": "Book a flight from SFO to Mumbai",
    "expected": {
        "agents": {"supervisor": {"output": "travel arrangements"},
                   "flight_agent": {"output": "flight booked"}},
        "evals": {"hallucination": "no_hallucination"},
    },
}

@pytest.mark.parametrize("testcase", [TESTCASE])
async def test_booking(monocle_trace_asserter, testcase):
    await monocle_trace_asserter.run_agent_async(root_agent, "google_adk", testcase=testcase)

    monocle_trace_asserter.called_agent(testcase=testcase).contains_output(testcase=testcase)
    monocle_trace_asserter.with_evaluation("okahu").check_eval(testcase=testcase)
```

Add an agent to the data and every one of those calls covers it.

It also adds `setup_test_cases()`, which turns what a workflow *already recorded* into that same shape — so a suite can be generated from production traces rather than written by hand.

## What's in it

**`testcase=` across the fluent API** — `run_agent`, `run_agent_async`, `with_trace_source`, `called_agent`, `called_tool`, `check_eval`, and the eight singular input/output assertions. Each reads only the part of the test case it needs.

**`setup_test_cases()`** builds cases from recorded runs, at any fact level:

| | |
|---|---|
| `source="okahu"` | enumerate a time window, read each fact's spans, optionally attach its eval labels |
| `source="local"` | load a committed JSON array — the model round-trips, so a discovered set can be frozen and replayed offline |
| `fact_name` above `traces` | a fact spans several traces; all their spans combine to describe it |
| `check_eval` | `True` for every eval, a name for one, `False` for none |
| `compare_eval` | borrow the expected label from a *different* eval — "run v2, expect what v1 recorded" |
| `eval_filter` | narrow which facts are considered at all |

**Three flows it enables** — eval regression (do recorded labels still reproduce?), eval tuning (does a new template match a golden one?), and A/B replay (does a live agent still behave like the recording?). All three are worked examples in the README.

**Supporting changes**: `FluentTestCase` accepts the shapes tests are actually written in (an `expected` wrapper, mapping-form collections, a scalar input) and rejects unknown keys; `Eval` gains an optional category via `{"name@category": "result"}`; `import_traces` gains a pure-fetch mode; the span loaders gain time-window and eval filters and a fact-ids lookup; parametrized cases are named after the test case instead of `testcase0`.

## Design decisions worth reviewing

- **Agents key on the name; tools key on the name *and* calling agent.** Two entries for one agent describe one agent and share a span list. One tool called by two agents is genuinely two span sets, so the caller is part of the key — which also means a test case naming the wrong caller now fails rather than matching anything.
- **Silence is not failure, but emptiness is.** An entry that sets no `input`/`output` is skipped, and a check where nothing sets one passes quietly — a test case describes what it knows. But a *selector* with nothing to select raises, so a vacuous suite can't read as green.
- **Two chain rules**, both enforced centrally so later work inherits them: a chain is all-`testcase=` or none of it, and a testcase chain uses one selector kind. Plain chains are untouched.
- **`get_test_cases` → `setup_test_cases`**, and it moved from `OkahuEval` to `OkahuSpanLoader` — it was mostly trace and span gathering, with the eval report as one optional step. No compat alias; the API is unreleased.

## Testing

780 unit tests pass, up from 619 at the start of this branch, with **no pre-existing failures left**: the branch also fixes 12 that were already broken — 9 for an unimplemented time-window feature whose tests were written first, and 3 from `MonocleValidator`'s process-global span state leaking between tests in one file.

Everything is stubbed at the HTTP boundary; no unit test touches the network. The Okahu response shapes are pinned against captured live responses.

## Two things a reviewer should know

**The `/evals/report` response shape was wrong on first implementation.** The label sits at `authoritative.eval_result.label` — one level deeper than the row. The original code read `authoritative.label`, which matched nothing and returned *zero* test cases for every real response, silently. Caught by pinning a real captured response; that payload is now a regression test.

**`setup_test_cases` is chatty.** One request per fact for its spans, another per fact above trace level to find its traces, plus one for the report. A wide window is a lot of calls. It's uncapped — freeze the result to JSON if you'll re-run it.

## Not in scope

`testcase=` on the aggregate selectors (`called_agents`, `called_tools`), the negative selectors (`does_not_call_*`), the scope assertions, and the eight `_any_` variants (they take any-of values, and an entry records exactly one input and one output).

Separately noticed and **not** fixed here: `called_agent(A).called_tool(T)` matches nothing in any trace. Narrowing is a flat span filter, so `called_agent` leaves only `agentic.invocation` spans and a following `called_tool` can never match an `agentic.tool.invocation` one. That contradicts the pattern CLAUDE.md documents, and `test_performance.py` asserts it. The working form is `called_tool(T, agent_name=A)`. Worth a separate fix to either the docs or the narrowing.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have signed the CLA

## Further comments


prasad-okahu:
@shambhavigoel-13, please take a look.

PR #810, opened 2026-09-02
Follow up fixes for parametric testing:

prasad-okahu:
## Proposed changes
Follow up fixes for parametric testing:
The eval-path bug — JSON has no Path type, so a string template path was sent as a built-in template name. This is the only true bug of the three.
Collection-time aborts — one purged trace meant zero tests ran, which reads as a broken module rather than one bad fact.
Timeout 30 → 120 + OKAHU_API_TIMEOUT — including why the defaults had to become None.

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments


cjames-okahu:
## Field verification of this PR against a live tenant

We built `monocle-test-tools` **0.8.15a2** from this branch (the published build's changed-file set matches this PR) and re-ran the evaluation we had done on 0.8.15a1, unchanged: 157 human-graded facts at fact level `traces`, six workflows, one 7-week window, a live Okahu tenant.

**All three fixes hold.**

**1. Collection-time aborts.** Previously one unloadable fact returned zero cases for its whole workflow, which took out two of our six workflows entirely. Now every workflow returns cases and the bad facts arrive carrying `load_error`:

| | 0.8.15a1 | 0.8.15a2 |
|---|---|---|
| workflows usable | 4 of 6 | **6 of 6** |
| worst workflow | aborts (0 cases) | 20 cases — 6 loaded, 14 carrying `load_error` |

The `setup_test_cases` tally (`N of M facts could not be loaded; each is reported as a failing test case`) is the right shape for a suite: the failures land in the failure column and the healthy facts still get graded.

**2. Timeout.** Verified that `OKAHU_API_TIMEOUT` actually reaches the socket rather than only the unit tests — against a non-routable address, where the elapsed time can only be the ceiling:

| `OKAHU_API_TIMEOUT` | result |
|---|---|
| `2` | `ConnectionError` after 2.01s |
| `5` | `ConnectionError` after 5.01s |
| `banana` | warns `is not a positive integer; using 120s`, proceeds |
| unset | 120 |

Making every `timeout` default `None` so the precedence is expressible is the right call — a numeric default would have been indistinguishable from a caller asking for that number.

**3. The eval-path bug.** This was the one that blocked us: a committed JSON case can only hold strings, so a custom template path was being sent as a built-in template *name*. Our reproduction from the 0.8.15a1 evaluation, unchanged:

```
test_door_testcase_from_json   PASSED     (FAILED on 0.8.15a1)
test_door_testcase_with_path   PASSED
```

**No regressions** in the consuming test kit: 162 passed / 1 skipped on both builds. We also ran a 30-case eval benchmark end to end on a2 (a custom template graded against human labels) — 30 passed, ~10 minutes, no library errors.

---

### Two things worth a look, both small and outside this PR's scope

**`OKAHU_API_TIMEOUT` only reaches `okahu_span_loader.py`.** The eval clients keep hardcoded values — `evals/okahu_eval.py` has `timeout=60` on the report pager (:129), the eval list (:281) and the fact map (:320), plus `timeout=120` (:550); `evals/okahu_filtered_eval.py` has four more. So span loading is now tunable and the report/eval-submission calls are not. Fine as scoped, just noting the other half.

**The sdist does not ship its unit-test fixtures.** Running the packaged tests gives 148 passed / 19 errors, and every error is `FileNotFoundError: tests/unit/traces/trace1.json`. Only affects anyone trying to run the tests from the package rather than the repo, so probably a line in the publish workflow rather than anything here.

### Separately, a defect in `okahu_span_loader.py` that this PR's fix uncovered

Not introduced here, and it was hidden by the aborts: `get_trace_ids` issues one GET on `/workflows/{workflow}/traces` with no pagination parameter, and `_unwrap_list` then discards the envelope — which is where `fact_count` and `next_page_token` live. `get_fact_ids` has the same gap.

At fact level `traces` the facts *are* the traces, so anything past the server's default page of 100 never becomes a fact at all: not a dropped case, not a `load_error` case, no log line. We hit it on a window holding 115 traces — 100 returned, `next_page_token` present and ignored, `fact_count` reporting 115, and four human-graded facts lost. Passing an explicit page size returns all 115.

Worth knowing that `evals/okahu_eval.py::_iter_eval_report_rows` already walks this exact protocol correctly, so it is a reference implementation rather than an open question. Happy to send a patch if useful.

shambhavigoel-13:
Looks Good , with testing outcomes and the feedback !

PR #813, opened 2026-09-04
Karmehr/okahu pagination

karmehr-okahu:
## Proposed changes

The Okahu `/traces` and `/facts/<name>/ids` endpoints paginate — each response
carries the page's payload, an advertised `fact_count`, and a `next_page_token`
when more remains. The span loader participated in none of it: `get_trace_ids`
and `get_fact_ids` each issued one GET, sent no `page_size`, and never sent a
token back. `_unwrap_list` then discarded the envelope those fields live in, so
the evidence that more data existed was destroyed before any caller could act
on it.

The result was a silent truncation at the server's `DEFAULT_PAGE_SIZE` of 100.
A trace past position 100 did not become a dropped case and did not become a
`load_error` case — it never entered the system. No log line, no tally, no
exception.

**This is invisible below 100 facts and hard-caps above it.** A window holding
58 traces was always correct, which is why it went unnoticed; a window holding
810 returned exactly 100. That cliff is what makes it P0 rather than a known
limit — discovery does not degrade gradually, it stops dead at 100 the moment a
window gets big enough.

Measured on stage, workflow `karmehr-okahu-monocle`, `2025-01-01 … 2026-09-03`,
running the pre-fix loader against the post-fix one over the same window:

| | before | after | server advertises |
|---|---:|---:|---:|
| `get_trace_ids` | 100 | 810 | 810 |
| `get_fact_ids("agent_requests")` | 100 | 809 | 809 |

### Review follow-up: a ceiling on how many facts are processed

Paginating removed an *accidental* cost ceiling — the old truncation at 100
bounded the work as a side effect. Discovery is now unbounded, so `max_facts`
bounds it deliberately: default 1000, overridable per call or with the
`OKAHU_MAX_FACTS` environment variable, which is the same parameter name,
variable and default `okahu_filtered_eval` already applies to a filtered run.

A window holding more **raises**, naming the count and the ceiling, rather than
returning the first 1000 — a silent subset is the defect this PR removes. Over
the same stage window: 810 traces passes, 3368 inferences raises.

The check lives in `setup_test_cases`, not in the enumerators. `get_trace_ids`,
`get_fact_ids`, `_collect_paged` and `_iter_pages` are byte-identical to the
pagination commits above, so `load_by_scope`, `import_traces` and
`from_okahu_scope` — which load one named session or scope and have no runaway
to bound — cannot inherit the ceiling. A signature-guard test enforces that.

The cap therefore counts facts (traces, agentic turns, inferences, agentic
sessions) and not the spans beneath a single fact, which is one level of
granularity below what it polices.

Out of scope, tracked separately: an empty window still yields zero cases, which
`parametrize` turns into a vacuously green run. That is pre-existing and
unaffected by this PR.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/docs/monocle_evaluation_api.md
+++ b/docs/monocle_evaluation_api.md
@@ -516,7 +516,7 @@ monocle_trace_asserter \
 |---|---|---|
 | `min_facts` | `1` | Fail the run if fewer than this many facts were discovered — guards against a vacuous pass on an empty window. |
 | `fail_threshold` | `0` | Allow up to this many non-matching facts before the assertion fails. |
-| `max_facts` | `OKAHU_MAX_FACTS` (`1000`) | Runaway guard — fail loudly if the window discovers more than this many facts. |
+| `max_facts` | `OKAHU_MAX_FACTS` (`1000`) | Runaway guard — fail loudly if the window discovers more than this many facts. Applies to filtered evals and to `setup_test_cases` discovery. |
 
 **Reading results.** Both the filtered and single-fact paths populate the same report (filtered = N facts; single-fact = a 1-fact report), on pass **and** failure:
 
--- a/test_tools/README.md
+++ b/test_tools/README.md
@@ -627,6 +627,8 @@ Each fact costs a request for its spans, plus one per fact above trace level to 
 
 Since a wide window now yields every fact rather than the first hundred, it also costs proportionally more collection time — roughly one span request per trace. Freeze to JSON, or narrow the window, if that matters to you.
 
+**Discovery is capped.** Because every page is now walked, a wide window can yield thousands of facts — and each one costs a span request plus, with `check_eval`, an eval. `max_facts` bounds that, defaulting to 1000 and overridable per call or with the `OKAHU_MAX_FACTS` environment variable (the same variable `okahu_filtered_eval` already honours). A window holding more raises, naming the count and the ceiling, rather than silently returning the first 1000 — a truncated subset is what the pagination fix set out to remove. Narrow the window, or raise the ceiling deliberately. The cap applies to `setup_test_cases` only: loading one named session or scope with `with_trace_source` is never bounded by it.
+
 **A fact that will not load does not cost you the window.** `setup_test_cases` runs at collection time, so raising would mean *no* tests run at all. Instead the fact is still returned, carrying the reason, and the first `testcase=` call it reaches fails with it — so it shows up as one failing test named after the fact while every other case runs normally. The bulk eval report is one call for all facts, so if *it* fails the reason is recorded against each of them.
 
 ```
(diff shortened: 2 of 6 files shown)


prasad-okahu (commented):
@karmehr-okahu, we should add a limit on number of traces/facts to process. The resultset for given time box could be very large and test will run forever and with potential large token cost for evals.
We should have a default of 1000 records which can be overridden by api argument and env variable.

cjames-okahu:
Independent verification of this branch against a **different tenant and endpoint** (prod, where the PR's own measurements are from stage), via the `0.8.15a3` patch wheel. The wheel's `okahu_span_loader.py` is byte-identical to head `f0e561f` after normalising its CRLF line endings, so what follows is this diff and nothing else.

Six workflows, one window, ~600 traces total. Before the fix every one of them returned exactly the server's first page; two of the windows hold more than that:

| | page 1, no page parameter | `fact_count` | this branch's `get_trace_ids` |
|---|---:|---:|---:|
| A | 100 | 100 | 100 |
| B | 100 | **115** | **115** |
| C | 100 | 100 | 100 |
| D | 100 | **101** | **101** |
| E | 100 | 100 | 100 |
| F | 100 | 100 | 100 |

Every one now reconciles with the advertised count, including the two that were silently short.

**The token walk is doing the work, not a larger page.** Workflow B at `page_size=50` (a forced three-page walk), at the new default of 200, and at 500 (a single page) all return the same 115 ids — same set *and* same order, no duplicates. That is also the direct evidence for the request-parameter naming note in `_iter_pages`: on these GET routes the token is accepted as `next_page_token`, and had the name been wrong the re-served page 1 would have surfaced as duplicates rather than as a short list.

`get_fact_ids` above trace level agrees: one workflow's `inferences` level returns 250 at `page_size=50` and 250 at `page_size=1000`, same set.

End to end, the cost this recovers for us: a discovery run that returned **23** test cases now collects **27** in 42s, and across the six workflows the human-graded facts discovery reaches went from **153** to **157 of 157**. The four it had been dropping all sat at positions 105–115 of the 115-trace window.

Both reconciliation nets fire, exercised offline against a stub `fetch` since the live server is well-behaved:

- advertises 115, serves 100, no token → returns 100 **and** logs `Okahu returned 100 of 115 advertised (…); the result is incomplete`
- echoes one token forever → stops after two fetches with `Okahu repeated a page token (…)`, rather than spinning at collection time
- `page_size` of `0`, `-1`, `1001`, `True`, `"200"`, `2.5` each raise `ValueError` naming the bound before any request goes out; `None` resolves to 200
- a bare-list envelope does not raise

Two notes, neither blocking:

1. **`_get_resource` re-resolves the apps/workflows namespace on every page.** The comment already says this is deliberate and costs nothing on the `apps`-correct path — agreed for a 3-page walk, worth a second look if page counts ever get large on a deployment that takes the fallback, since it is one wasted 404 per page.

karmehr-okahu:
> @karmehr-okahu, we should add a limit on number of traces/facts to process. The resultset for given time box could be very large and test will run forever and with potential large token cost for evals. We should have a default of 1000 records which can be overridden by api argument and env variable.

Just checking, do we want to keep the default max facts separate from max spans. Currently they are conflated in Monocle so I just want to make sure it is correct for me to separate and only assert on the max facts/traces rather than max spans.

prasad-okahu:
> > @karmehr-okahu, we should add a limit on number of traces/facts to process. The resultset for given time box could be very large and test will run forever and with potential large token cost for evals. We should have a default of 1000 records which can be overridden by api argument and env variable.
> 
> Just checking, do we want to keep the default max facts separate from max spans. Currently they are conflated in Monocle so I just want to make sure it is correct for me to separate and only assert on the max facts/traces rather than max spans.

Good point. I was referring to default max facts which I think is a bigger concern.

prasad-okahu:
> > > @karmehr-okahu, we should add a limit on number of traces/facts to process. The resultset for given time box could be very large and test will run forever and with potential large token cost for evals. We should have a default of 1000 records which can be overridden by api argument and env variable.
> > 
> > 
> > Just checking, do we want to keep the default max facts separate from max spans. Currently they are conflated in Monocle so I just want to make sure it is correct for me to separate and only assert on the max facts/traces rather than max spans.
> 
> Good point. I was referring to default max facts which I think is a bigger concern.

@karmehr-okahu, I'll go ahead and merge the patch. Please add the factID limit as a follow up patch.

prasad-okahu (approved):

karmehr-okahu:
> > > > @karmehr-okahu, we should add a limit on number of traces/facts to process. The resultset for given time box could be very large and test will run forever and with potential large token cost for evals. We should have a default of 1000 records which can be overridden by api argument and env variable.
> > > 
> > > 
> > > Just checking, do we want to keep the default max facts separate from max spans. Currently they are conflated in Monocle so I just want to make sure it is correct for me to separate and only assert on the max facts/traces rather than max spans.
> > 
> > 
> > Good point. I was referring to default max facts which I think is a bigger concern.
> 
> @karmehr-okahu, I'll go ahead and merge the patch. Please add the factID limit as a follow up patch.

I added the max facts into this PR as of friday. So it should already be merged then

PR #814, opened 2026-09-08
fix(test_tools): repair integration test failures

shambhavigoel-13:
Fixes a set of unrelated defects that together accounted for most of the integration-suite failures.

Product fixes:
- validator: post_test_cleanup already stops the scope token, but both wrappers stopped it a second time. The double detach raised "RuntimeError: <Token ...> has already been used once" from the finally block while the real exception was propagating, burying every failing test's actual error.
- pytest_plugin: honour @pytest.mark.xfail for fluent assertions. They record rather than raise, so the outcome is flipped to "failed" after pytest's own xfail hookwrapper has already seen a passing call, leaving the marker unapplied. ~26 intentional-failure tests reported as hard FAILED; three that silently stopped failing were masked as passes.
- llamaindex_runner: only pass memory= to callables that accept it, and bind it with functools.partial for run_in_executor, which takes no kwargs.
- instrumentor: reset_span_processors now re-appends the trace-return processor like setup_monocle_telemetry does. Rebuilding the processor list on an initialised instrumentor silently dropped trace-return for the rest of the process.

Test fixes:
- crewai: use the parametrized case rather than the module-level list; drop an await on a sync helper; run_async_agent -> run_agent_async.
- llamaindex: migrate to llama-index 0.13 (ReActAgent.from_tools removed, chat/achat/aquery replaced by run()); extract text from AgentOutput so the ChatMessage role prefix does not break the similarity comparer.
- adk assertions: tool input is serialized JSON with per-run key order, so the exact-match has_*/does_not_have_* forms can never match a fragment. Switched five call sites to the substring forms.
- okahu trace assertions: make the pinned workflow/trace id overridable and skip on 401/403/404 instead of failing on absent tenant data.
- http trace return: stop double-registering the pytest plugin already registered by conftest, and enable trace-return in conftest so it is set before the first setup_monocle_telemetry() regardless of import order.

Verified: apptrace unit suite 855 passed; integration failures 77 -> 15, with the remainder being Okahu tenant configuration, tuned-too-tight performance thresholds, and LLM nondeterminism.

## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments

### DIFF
--- a/apptrace/src/monocle_apptrace/instrumentation/common/instrumentor.py
+++ b/apptrace/src/monocle_apptrace/instrumentation/common/instrumentor.py
@@ -406,6 +406,9 @@ def setup_monocle_telemetry(
     return get_monocle_instrumentor()
 
 def reset_span_processors(span_processors:list[SpanProcessor]):
+    # Mirror setup_monocle_telemetry: rebuilding the processor list without this
+    # silently drops trace-return for the rest of the process.
+    span_processors = _append_trace_return_processor(span_processors)
     monocle_span_processor = get_monocle_span_processor()
     if monocle_span_processor:
         clear = getattr(monocle_span_processor, "clear_span_processors", None)
--- a/test_tools/src/monocle_test_tools/pytest_plugin.py
+++ b/test_tools/src/monocle_test_tools/pytest_plugin.py
@@ -150,6 +150,29 @@ def pytest_runtest_makereport(item, call):
             rep.outcome = "failed"
 
             rep.longrepr = traceAssertion.get_assertion_messages()
+            _apply_xfail_to_deferred_failure(item, rep)
+
+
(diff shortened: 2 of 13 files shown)


prasad-okahu (approved):
@shambhavigoel-13 , changes look fine. Please update the CDO for all commits in the branch.

prasad-okahu:
@shambhavigoel-13 , the DCO is missing for the PR. Please follow the steps here to update - https://github.com/monocle2ai/monocle/pull/814/checks?check_run_id=102186602709

shambhavigoel-13:
cc:- @prasad-okahu

PR #819, opened 2026-09-10
fix(test_tools): report failed trace exports and empty eval results accurately

### LINKED ISSUE #816: test_tools: eval backend failures are silent or misreported as assertion failures
## Summary

When the evaluation backend is unreachable, or accepts a request but returns no results, `test_tools` reports the condition as a **content assertion failure** rather than an infrastructure problem. In a recent CI run this caused a complete trace-export outage to surface on 14 tests as *"Expected 'result' key in response"* — a message describing the wrong problem entirely.

Four independent defects contribute. Each is separately fixable; items 1–3 are small and unambiguous, item 4 needs a design decision.

---

### 1. `export_trace()` discards the export result and claims success ⭐ highest value

`test_tools/src/monocle_test_tools/evals/okahu_eval.py:258-263`

```python
exporter = okahu_exporter.OkahuSpanExporter(evaluate=True)
exporter.export(filtered_spans)      # ← return value discarded
exporter.shutdown()
self._trace_exported = True          # ← set unconditionally
```

The exporter correctly detects failure and returns `SpanExportResult.FAILURE` on any non-2xx (`apptrace/src/monocle_apptrace/exporters/okahu/okahu_exporter.py:95-100`). The caller ignores it and then marks the trace as exported. Every subsequent eval queries a backend that was never written to, and reports a content failure.

**Fix:** check the return value; raise `AssertionError` naming the status code when it is `FAILURE`, and leave `_trace_exported` as `False`.

---

### 2. An empty result is misreported as a malformed response

`okahu_eval.py:571-584`

```python
eval_result = data.get("result") or []
parsed = json.loads(eval_result[0].get("result"))   # IndexError when []
...
except Exception as exc:
    raise AssertionError(
        f"Unexpected response format from evaluation service. Expected 'result' key in response. ..."
    )
```

An `IndexError` on an empty list is reported as *"Expected 'result' key in response"* — but the key **is** present and simply empty. The message sends the reader toward response parsing when the actual condition is "the backend returned no results for this fact."

**Fix:** test `if not eval_result:` before indexing and raise a distinct, accurate message. Narrow the bare `except Exception` so genuine format errors stay distinguishable.

Note: the same pattern and message appear twice — at `okahu_eval.py:450-458` and `:571-584`. Both need the fix.

#### Why this is not subsumed by item 1

Item 1 removes the most common *cause* of an empty result; this item fixes how an empty result is *reported*, whatever the cause. Fixing 1 does not short-circuit 2, because there are at least three routes to an empty result where item 1's check never fires:

1. **Replay mode skips the export path entirely.** `okahu_eval.py:70` sets `self._trace_exported = self._trace_source == "okahu"`, and `:495` only calls `export_trace()` when that is `False`. For replay-sourced tests, `export_trace()` is never invoked, so item 1's guard is inert and this message is the only thing the user sees.
2. **Ingestion is asynchronous.** The ingest endpoint returns `202 Accepted`; the eval submits milliseconds later. An indexing lag yields an empty result from a genuinely *successful* export.
3. **Server-side scoping can filter everything out.** Traces and spans can be retrieved successfully and then dropped by the eval template's span-type/role scoping, producing an empty result with no export failure at all.

Other causes in the same class: wrong `fact_name`, time-window mismatch, workflow-name mismatch, or facts since purged. In every one of these the export succeeded, so only this item's fix improves the message.

---

### 3. Environment variable handling is inconsistent and unsafe

**3a — an empty variable silently defeats the default** (lines 275, 315, 489, 621):

```python
base = os.getenv("OKAHU_EVALUATION_ENDPOINT", DEFAULT_ENDPOINT).rstrip("/")
```

`os.getenv(name, default)` returns `""` — not the default — when the variable is *set but empty*, yielding `base = ""` and a schemeless URL. This has already produced a real CI failure (`Invalid URL '/v1/eval/templates': No scheme supplied`).

**Fix:** `os.getenv(name) or DEFAULT`.

**3b — `.strip()` on a possibly-`None` value** (lines 254, 271):

```python
api_key = (os.getenv("OKAHU_API_KEY")).strip()
```

Raises `AttributeError: 'NoneType' object has no attribute 'strip'` instead of the clear `AssertionError("OKAHU_API_KEY is not configured.")` defined two lines below. Lines 311, 476, and 616 already use the safe `(os.getenv(...) or "")` form — the file is internally inconsistent.

**Fix:** apply the safe form uniformly.

---

### 4. 🔶 Infrastructure failures are indistinguishable from assertion failures — *design decision required*

`test_tools/src/monocle_test_tools/fluent_api.py:240-254`

`record_assertion()` collapses every condition into `AssertionError` + a string, so a genuine mismatch (label was `negative`, expected `positive`) and an infrastructure problem (backend unreachable, nothing ingested, HTTP 5xx) both surface identically as pytest `FAILED`.

Reporting *"hallucination check failed"* when the truth is *"the backend was unreachable"* is the root diagnostic problem behind items 1–2.

**This item is not ready to implement — it needs a decision on outcome mapping.** Straw proposal:

| Condition | Outcome |
|---|---|
| Export failed / backend unreachable / 5xx | ERROR |
| Required env or credentials absent | SKIP |
| Eval ran, label ≠ expected | FAILED |
| Eval ran, returned no results | ERROR (data problem, not a user assertion) |

The last row is the debatable one. Open questions: should a no-results eval be ERROR or FAILED? Does introducing a non-`AssertionError` path break existing consumers of the fluent API?

---

## Acceptance criteria (items 1–3)

- [ ] A failed trace export raises immediately, naming the status code; `_trace_exported` stays `False`.
- [ ] An empty result produces a distinct message that does not claim the `result` key is missing.
- [ ] All environment reads use `or DEFAULT`; no `.strip()` on a possibly-`None` value.
- [ ] Tests cover: export failure; empty result; empty-string environment variable; missing API key.

## Sequencing

Items 1–3 are independent and each worth shipping alone — item 1 closes most of the diagnostic gap by itself. Item 4 should be agreed before any implementation.

L4XB:
## Proposed changes

Fixes #816 (items 1–3).

`OkahuEval.export_trace` discarded the exporter's return value, so a failed upload (bad API key, 5xx, timeout) still marked the trace as exported and every later eval queried a backend that had nothing. That surfaced much later as `Unexpected response format ... Expected 'result' key in response`, which also fired for a perfectly well-formed response whose `result` list was simply empty.

- `export_trace` now shuts the exporter down in a `finally`, raises when nothing was uploaded (every span filtered out), and raises with the HTTP status (or "no HTTP response (timeout or exporter shut down)") when the export failed. `_trace_exported` stays `False`.
- `OkahuSpanExporter` records `last_status_code` for the most recent upload so the eval can name the status. (Small addition in apptrace; no behaviour change there.)
- Both result-parsing sites treat an empty `result` list as "no results for this fact" (with the job id and the likely causes: trace not ingested/indexed yet, fact or workflow name mismatch, template scoping dropped every span) instead of a format error. The format error itself now says what shape was expected (`result[0].result` as a JSON string with `label`/`explanation`).
- Item 3: `OKAHU_EVAL_TIME_PAD_SECONDS` falls back to the default when the variable is set but empty, matching the other env reads that already use `or`.

Item 4 (mapping infrastructure failures to ERROR/SKIP instead of `AssertionError`) is deliberately not touched: the issue asks for a design decision first. All raises here stay `AssertionError`, so existing consumers of the fluent API are unaffected. The `.strip()` on a possibly-`None` key mentioned under item 3 is already safe on `main` (`(os.getenv(...) or "").strip()`).

## Types of changes

## Checklist

- [x] I have signed the CLA (DCO sign-off on the commit)
- [x] Lint and unit tests pass locally with my changes (`pylint` on the two touched files goes from 8.85 to 8.95, one `invalid-envvar-default` warning removed; `test_tools/tests/unit` okahu eval suites: 38 passed)
- [x] I have added tests that prove my fix is effective or that my feature works (`test_tools/tests/unit/test_okahu_eval_export_and_empty_result.py`: failed export with/without HTTP status, nothing uploaded, successful export, empty `result`, malformed entry still a format error, empty pad env, missing API key)
- [ ] I have added the necessary documentation (if appropriate) — not needed, error messages only
- [x] Any dependent changes have been merged and published in downstream modules — none

## Further comments

The `last_status_code` attribute is read with `getattr(..., None)` so a test double or an older exporter without it still produces a useful message.

### DIFF
--- a/apptrace/src/monocle_apptrace/exporters/okahu/okahu_exporter.py
+++ b/apptrace/src/monocle_apptrace/exporters/okahu/okahu_exporter.py
@@ -40,6 +40,9 @@ class OkahuSpanExporter(SpanExporterBase):
         self.endpoint = endpoint or okahu_endpoint
         api_key: Optional[str] = _get_okahu_api_key()
         self._closed = False
+        # HTTP status of the most recent synchronous upload, so a caller that
+        # gets SpanExportResult.FAILURE can say why (None: no response, e.g. timeout).
+        self.last_status_code: Optional[int] = None
         if not api_key:
             raise ValueError("OKAHU_API_KEY not set.")
         self.timeout = timeout or 15
@@ -81,11 +84,13 @@ class OkahuSpanExporter(SpanExporterBase):
 
         def send_spans_to_okahu(span_list_local=None, is_root=False):
             try:
+                self.last_status_code = None
                 result = self.session.post(
                     url=self.endpoint,
                     data=json.dumps(span_list_local),
                     timeout=self.timeout,
                 )
+                self.last_status_code = result.status_code
                 if result.status_code not in REQUESTS_SUCCESS_STATUS_CODES:
                     logger.error(
                         "Traces cannot be uploaded; status code: %s, message %s",
--- a/test_tools/src/monocle_test_tools/evals/okahu_eval.py
+++ b/test_tools/src/monocle_test_tools/evals/okahu_eval.py
@@ -3,6 +3,7 @@ import logging
 import os
 from datetime import datetime, timezone
 from opentelemetry.sdk.trace import Span
+from opentelemetry.sdk.trace.export import SpanExportResult
 import requests
 from opentelemetry.baggage.propagation import W3CBaggagePropagator
 from monocle_apptrace.exporters.okahu import okahu_exporter
@@ -77,6 +78,22 @@ class OkahuEval(BaseEval):
         """
         from monocle_test_tools.evals.okahu_eval_discovery import discover_fact_evals as _discover
         return _discover(spans, fact_name=fact_name)
+
+    @staticmethod
+    def _no_results_message(job_id, data) -> str:
+        """An empty ``result`` list is a data condition, not a malformed response."""
+        return (
+            f"Evaluation service returned no results for this fact (job {job_id}). The trace may not be "
+            "ingested or indexed yet, the fact or workflow name may not match, or the eval template's "
+            f"scoping dropped every span. Received: {data}"
+        )
+
+    @staticmethod
+    def _unexpected_format_message(data) -> str:
+        return (
+            "Unexpected response format from evaluation service: expected result[0].result to be a JSON "
+            f"string carrying 'label' and 'explanation'. Received: {data}"
+        )
 
     @classmethod
     def _eval_report_by_fact(cls, *, workflow_name, fact_ids, fact_name, start_time,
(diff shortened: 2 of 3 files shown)


prasad-okahu:
@L4XB Thanks for the contribution.
@karmehr-okahu , could you please take a look and add your feedback.

karmehr-okahu:
@prasad-okahu Everything looks good. I reviewed the code in respect to the ticket and it correctly addresses the issues 
@L4XB  Thanks for the contribution!

prasad-okahu (approved):

prasad-okahu:
@L4XB Merged PR to main. Thanks for your contribution!

